PR #28, opened 2014-11-27
WIP autoproj versioning functionality

doudou:
IMPORTANT: this requires the misc_cleanup branch of autobuild to be present. I'm trying to get it reviewed and merged bit-by-bit ...

This pull request implements the support in autoproj for versioning, that is the ability to replicate an installation 1:1 without having to fear for changing software or package sets.

The following subcommands are planned:
 - creation of autoproj/overrides.d in which versioning files are saved
 - autoproj versions: generates a version file in which all packages are "pinned"
 - autoproj commit: generates a version file within the current autoproj configuration and commit it (assumes that autoproj/ is under git)
 - autoproj tag: generates a version file, and creates a commit+tag in the current autoproj/ git repository but leaves autoproj/ unchanged
 - auto-log: saves a log of all the updates done to the installation
 - autoproj log: displays the log
 - autoproj reset: resets the install to a known state from the log


jakobs:
I don't have a problem in general with working on master. However, I am currently a bit confused to the state of the branches. There are a lot of different branches for e.g. autobuild. Which one is the right one to use here? Also, it would be nice to discuss and describe the functionality of the new commands in a wiki.

doudou:
and rebased on current master

jakobs:
autoproj commit does not work with bash and sh:
the autoproj commit script requires the acd shell function. However, its not exported in bash, and it seems it cannot be imported for sh.

jakobs:
I also had another bug related to the cachinfo. Created a pull request for that (#33)

doudou:
> autoproj commit does not work with bash and sh

That acd is not available on bash and sh would be a bug in itself. I have some bash machines here where it works, though, so I don't know what happens.

In any case, autoproj commit should definitely NOT depend on acd being present.

doudou:
I've rebased the branch on the new master (I've merged some smaller PRs that were sitting around for a month or so). I had also to rebase the improve_handling_of_detached_head branch on autobuild. These two branches are all that we need now

PR #33, opened 2015-01-09
fix: added missing splat operator to cacheinfo parameters

jakobs:
in case ruby version is less than 2.1.0, the cachinfo parameters have to
presented as individual arguments. Passing the array directly resulted
in a bad syntax for the git command line.


doudou:
Merged. I think it was even a bug-in-waiting on 2.1

PR #52, opened 2015-03-10
os-detection: use lsb_release also for debians

marvin2k:
On my _Debian Jessie_, the os-autodetection returns `debian, 8.0`, which is wrong in the way that our osdeps use `jessie` or `sid` -- not the release-number.

The file `/etc/debian_version` does not contain something like `sid` on a current _debian-unstable_, only the actual release number like `8.0`. But I'm not sure if this changes once Debian leaves the current feature-freeze...

Well, this is my shot.

It could be argued that it is unfeasible to track a moving target like `sid` in the osdeps. Also, unstable is not really officially supported...?

Signed-off-by: Martin Zenzes <martin.zenzes@dfki.de>

### DIFF
--- a/lib/autoproj/osdeps.rb
+++ b/lib/autoproj/osdeps.rb
@@ -1276,11 +1276,24 @@ fi
             Autobuild.progress :operating_system_autodetection, "autodetecting the operating system"
             names, versions = os_from_os_release
 
-            # Don't use the os-release information on Debian, since they
-            # refuse to put enough information to detect 'unstable'
-            # reliably. So, we use the heuristic method for it
-            if !names || names[0] == 'debian'
+            if !names
                 names, versions = guess_operating_system
+            end
+
+            # on Debian, they refuse to put enough information to detect
+            # 'unstable' reliably. So, we use the heuristic method for it
+            if names[0] = "debian"
+                # check if we actually got a debian with the "unstable" (sid)
+                # flavour. it seems that "/etc/debian_version" does not contain
+                # "sid" (but "8.0" for example) during the feature freeze
+                # phase...
+                if File.exists?('/etc/debian_version')
+                    debian_versions = [File.read('/etc/debian_version').strip]
+                    if debian_versions.first =~ /sid/
+                        versions = ["unstable", "sid"]
+                    end
+                end
+                # otherwise "versions" contains the result it previously had
             end
             return if !names
 
--- a/lib/autoproj/osdeps.rb
+++ b/lib/autoproj/osdeps.rb
@@ -1322,8 +1322,7 @@ fi
                 return
             end
 
-            distributor = `lsb_release -i -s`
-            distributor = distributor.strip.downcase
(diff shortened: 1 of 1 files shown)


doudou:
Debian is extremely annoying in that they don't put the 'unstable' information in /etc/os-release

But you are right in pointing out that it does contain useful information. I would prefer: using os-release and then adding 'unstable' and 'sid' based on what is in /etc/debian_version

The "official support" status has nothing to do with autoproj. autoproj is much more generic than Rock. Crippling it just because Rock cannot support debian officially sounds just like a bad idea. IMO, autoproj should be able to detect testing/unstable as it *is* what most Debian-using people use, there is a reliable way to detect it (through /etc/debian_version outside of feature freeze AFAIK) *and* it worked just fine so far.

marvin2k:
I constantly forget that autoproj is intended to not only be used in rock....

So:
- Why does autoproj needs to detect `testing`? Isn't the release name `jessie` enough?
- Where to find information on _what_ is actually supposed to be inside `/etc/debian_version`? For all the source-archives for `base-files` I find on [packages.debian.org](https://packages.debian.org/eo/source/jessie/i386/base-files) this file contains the version _number_. Which is `8.0` for `jessie` and `sid`...

doudou:
> Why does autoproj needs to detect testing? Isn't the release name jessie enough?

Autoproj does not, really. We could move to a scheme where the current sid/unstable is actually flagged as "the next release name". However, we need some time to move to there, because quite a lot of people's osdeps depend on testing and/or sid to function.

marvin2k:
Sorry, I don't get it...

- Using `testing` does not make sense since it's meaning is changing over time. Also, most (all?) osdeps use `jessie`
- Currently missing is reliable detection of `unstable`/`sid`. Using it does not really make that much sense since the Debian package-set it represents is changing over time.
- Anyone who relies on the string `sid` to work is f*d already, since autoproj happily reports `8.0` instead.

doudou:
> Using testing does not make sense since it's meaning is changing over time. Also, most (all?) osdeps use jessie

Maybe it started to be so recently, but I know for a fact that it was not the case for a while. Meaning: there are probably builds that depend on 'testing' or 'unstable' being there.

> Using it does not really make that much sense since the Debian package-set it represents is changing over time.

Not that much actually. It is how it worked the last 5 years or so ... In addition, not supporting unstable at all will break any unstable user every once in a while for a few weeks *with no simple fix possible* since new packages arrive in unstable before testing.

> Anyone who relies on the string sid to work is f*d already, since autoproj happily reports 8.0 instead.

Let's fix that, instead of saying "autoproj is currently broken, let's make it officially broken"

In any case, using os-release *and* checking for 'sid' in /etc/debian_version would fix both your problem and keep the current behaviour. So what's the problem ?

marvin2k:
My understand of the problem is the problem ;-) And the commit represents my attempt to scratch my itch. I'll update it.

PR #56, opened 2015-03-18
add explicit support for renaming packages

doudou:
This adds support for renaming packages explicitely instead of
directly accessing the metapackage creation. Moreover, it also
adds support to obsolete an old name (generating an error).


goldhoorn:
ack

doudou:
Actually, this is do-not-merge because of #58. The way to force-building an excluded package had unintended consequences, and had to be removed. We'll have to figure another one out before we can merge this.

PR #62, opened 2015-03-30
Handle non git main configurations in update log (fixes #61)

### LINKED ISSUE #61: autoproj-1.13.1: current_import_state': undefined method `show'
calling autoproj update results in the following error message:

autoproj-1.13.1/lib/autoproj/ops/snapshot.rb:169:in current_import_state': undefined methodshow' for nil:NilClass (NoMethodError)

I looked into the code and I think it cannot be assumed that main.importer is returning an object.

The vhs which has no importer is (the git is a clone from another git on the hard disk):
local:/var/lib/jenkins/workspace/ropencv_gem/dev/autoproj

same propblem here:
gems/autoproj-1.13.1/lib/autoproj/ops/snapshot.rb:191:in save_import_state': undefined methodgit_dir' for nil:NilClass (NoMethodError)

doudou:
This properly handles main configurations that are either not a git, or are a git repository but that is not declared in config.yml

### DIFF
--- a/lib/autoproj/cmdline.rb
+++ b/lib/autoproj/cmdline.rb
@@ -773,13 +773,12 @@ module Autoproj
             return all_enabled_packages
 
         ensure
-            if !updated_packages.empty?
+            if Autoproj.config.import_log_enabled? && !updated_packages.empty?
                 failure_message =
                     if $!
                         " (#{$!.message.split("\n").first})"
                     end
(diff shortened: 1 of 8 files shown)


doudou:
Can you try it again ? The PR contained code from the master branch that I don't feel confident releasing right now ... I've rebased.

PR #66, opened 2015-04-28
Load in overrides.rb directory also the .rb files

goldhoorn:


goldhoorn:
I assume this is the wrong place & time but need hint for solving

doudou:
I have no problem with this, but (1) they should be loaded *after* the package sets and (2) I am afraid I would prefer having this done against the autoproj-2.0 branch (I did a major refactoring of the autoproj command-line interface)

goldhoorn:
superseeded by #67

PR #67, opened 2015-04-28
Load in overrides.rb directory also the .rb files

goldhoorn:
...old pr #66

### DIFF
--- a/lib/autoproj/workspace.rb
+++ b/lib/autoproj/workspace.rb
@@ -435,6 +435,10 @@ module Autoproj
             # Resolve optional dependencies
             manifest.resolve_optional_dependencies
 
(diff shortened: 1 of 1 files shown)


lib/autoproj/workspace.rb
```diff
@@ -435,6 +435,10 @@ def finalize_package_setup
             # Resolve optional dependencies
             manifest.resolve_optional_dependencies
 
+            Dir.glob(File.join( Autoproj.overrides_dir, "*.rb" ) ).sort.each do |file|
+                require file
```
doudou: Use Workspace#load to load the files.

goldhoorn:
*up*

PR #71, opened 2015-05-19
added commandline option to build all known packages

goldhoorn:


marvin2k:
To provide some context: This was deemed to be needed for debugging or testing the build-process. The build-server always wants to build all known packages for example. Maintaining an addtional "all-buildconf" is not a nice solution in the long-term. Having this option in autoproj might be a nice way to do this?

goldhoorn:
I merge because this starts to become a blocker for the release and needs to be tested on a larger setup.

PR #73, opened 2015-05-22
Automatic rock-release PR: integrate rc-patches in master

goldhoorn:

PR #75, opened 2015-05-26
merge rock-rc in master (only adds all-known-packages)

goldhoorn:
This superseeds #73 
All non-required commits are reverted so that a clean git-tree can be created


doudou:
The non-clean commits never landed on anything anyone should be using (since autoproj is not shipped through git but through rubygems), so please remove them completely.

goldhoorn:
*updated*

PR #79, opened 2015-06-16
Fix reporting on missing packages to account for aliases

2maz:
Error reporting did not report the actually missing package name, but only the high level alias, which is quite misleading to fixing the error.

### DIFF
--- a/lib/autoproj/osdeps.rb
+++ b/lib/autoproj/osdeps.rb
@@ -1357,6 +1357,20 @@ fi
 
         class InvalidRecursiveStatement < Autobuild::Exception; end
 
+        # Return the path to the osdeps name for a given package name while
+        # accounting for package aliases
+        #
+        # returns an array contain the path starting with name and
+        # ending at the resolved name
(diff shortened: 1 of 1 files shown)


doudou:
You should definitely report the whole chain. Otherwise, one will wonder why this osdep is required in the first place (since the names do not match the dependency).

Btw, how are you using these aliases (I even wonder what they are used for ...) ?

2maz:
> You should definitely report the whole chain. Otherwise, one will wonder why this osdep is required in the first place (since the names do not match the dependency).
sure can update accordingly

> Btw, how are you using these aliases (I even wonder what they are used for ...) ?
just trying to improve the current error reporting since autoproj was complaining about the missing dependency for ruby while it actually meant ruby22 in that particular case.
Otherwise it has been your commit  fd19cf and seems it is specifically for ruby.

doudou:
> Otherwise it has been your commit fd19cf and seems it is specifically for ruby.

Indeed. Thanks for taking the time to explain my code to me ;-)

PR #81, opened 2015-06-30
Make repository id equal to the url it is referring to

2maz:
Otherwise multiple configurations (e.g. to legacy naming in package_sets) which refer to the same url lead
to broken links under autoproj/remotes/.

This might not be the best fix for the given issue so @doudou let me know in case there is a better place to fix this.
In  the current situation there are two such definitions, but only one package_set is found under .remotes and the symlink points to the not-existing one.

### DIFF
--- a/lib/autoproj/gitorious.rb
+++ b/lib/autoproj/gitorious.rb
@@ -93,7 +93,7 @@ module Autoproj
                  url: "#{pull_base_url}#{url}",
                  push_to: "#{push_base_url}#{url}",
                  retry_count: 10,
-                 repository_id: "#{name.downcase}:#{url}"].merge(vcs_options)
+                 repository_id: "#{url}"].merge(vcs_options)
         end
     end
 end


doudou:
This looks very wrong. You would end up having e.g. github:rock-drivers/package_set and gitlab_internal:rock-drivers/package_set be the same configuration.

Can you give a more specific example ?

2maz:
Hope the following clarifies the situation:
- while autoproj allows for multiple configurations of a gitorious server, the link under remotes is generated based on the package_set name and since they refer to the same package_set it will be only one link -- which is fine as long as the url is the same (that is why the fix works)
- multiple package_sets are to be expected under .remotes when multiple such configurations exist, e.g. github_rock_drivers or gitlab_internal_rock_drivers, but there is just one persisting
- eventually the symlink from remotes/rock.drivers -> .remotes/gitlab_internal_rock_drivers which in this case does not exist anymore  leads to a broken link

doudou:
If two package sets with the same name are imported, autoproj should (1) pick one and only one, (2) warn about the duplicate and (3) keep both in .remotes (to avoid re-checking out unnecessarily). There seem to be a bug there in at least (3).

Ops::Configuration#create_remote_set_user_dir is called unconditionally after the package set got updated (https://github.com/rock-core/autoproj/blob/master/lib/autoproj/ops/configuration.rb#L232), which updates the symlink, but since the first package set encountered is kept, the repository is removed at cleanup (https://github.com/rock-core/autoproj/blob/master/lib/autoproj/ops/configuration.rb#L270). So, double-bug. Luckily for us, they did not cancel each other out.

So, first fix is to move the call to create_remote_set_user_dir after the already_loaded test (https://github.com/rock-core/autoproj/blob/master/lib/autoproj/ops/configuration.rb#L236).

The second fix is to store the list of needed remotes_dir and user_dir in the main loop of #load_update_package_sets and feed that to the two cleanup methods, instead of relying on package_sets. It would avoid deletion of the duplicate package set (which is good) and make the determination of what should or should not be kept straigtforward.

2maz:
K, opened #83 for that, where we can discuss the details further if needed. 
Closing this one.

PR #83, opened 2015-07-03
Fix handling of redundant package set definitions

2maz:
The user will be warned about a redundant package set definition and informed which will be the one kept.
Meanwhile, all package sets which have been checked out during the update process will remain in the .remotes folder.
Other (not handled) package sets e.g. after changing the configuration will be removed.

### DIFF
--- a/lib/autoproj/ops/configuration.rb
+++ b/lib/autoproj/ops/configuration.rb
@@ -209,6 +209,9 @@ module Autoproj
             by_repository_id = Hash.new
             by_name = Hash.new
 
+            required_remotes_dirs = Array.new
+            required_user_dirs = Array.new
+
             queue = queue_auto_imports_if_needed(Array.new, root_pkg_set, root_pkg_set)
             while !queue.empty?
                 vcs, options, imported_from = queue.shift
@@ -227,17 +230,28 @@ module Autoproj
                 end
                 by_repository_id[repository_id] = [vcs, imported_from]
 
+                # Make sure the package set has been already checked out to
+                # retrieve the actual name of the package set
                 if !vcs.local?
                     update_remote_package_set(vcs, only_local)
-                    create_remote_set_user_dir(vcs)
+                    raw_local_dir = PackageSet.raw_local_dir_of(vcs)
+                    required_remotes_dirs << raw_local_dir
                 end
-
                 name = PackageSet.name_of(manifest, vcs)
+
+                required_user_dirs = by_name.collect { |k,v| k }
+                Autoproj.debug "Trying to load package_set: #{name} from definition #{repository_id}"
+                Autoproj.debug "Already loaded package_sets are: #{required_user_dirs}"
+
                 if already_loaded = by_name[name]
                     already_loaded_pkg_set, already_loaded_vcs = *already_loaded
                     if already_loaded_vcs != vcs
                         if imported_from
-                            Autoproj.warn "#{imported_from.name} auto-imports a package set from #{vcs}, but a package set with the same name (#{name}) has already been imported from #{already_loaded_vcs}, I am skipping this one"
+                            Autoproj.warn "redundant auto-import by #{imported_from.name} for package set '#{name}'."
+                            Autoproj.warn "    A package set with the same name (#{name}) has already been imported from"
+                            Autoproj.warn "        #{already_loaded_vcs}"
+                            Autoproj.warn "    Skipping the following one: "
+                            Autoproj.warn "        #{vcs}"
                         else
                             Autoproj.warn "the manifest refers to a package set from #{vcs}, but a package set with the same name (#{name}) has already been imported from #{already_loaded_vcs}, I am skipping this one"
                         end
@@ -248,6 +262,8 @@ module Autoproj
                         imported_from.imports << already_loaded_pkg_set
                     end
                     next
+                else
+                    create_remote_set_user_dir(vcs)
                 end
 
                 pkg_set = load_package_set(vcs, options, imported_from)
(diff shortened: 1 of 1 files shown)


lib/autoproj/ops/configuration.rb
```diff
@@ -227,17 +242,30 @@ def load_and_update_package_sets(root_pkg_set, only_local = false)
...
+
                 if already_loaded = by_name[name]
                     already_loaded_pkg_set, already_loaded_vcs = *already_loaded
                     if already_loaded_vcs != vcs
                         if imported_from
-                            Autoproj.warn "#{imported_from.name} auto-imports a package set from #{vcs}, but a package set with the same name (#{name}) has already been imported from #{already_loaded_vcs}, I am skipping this one"
+                            Autoproj.warn "redundant auto-import by #{imported_from.name} for package set '#{name}'."
+                            Autoproj.warn "    A package set with the same name (#{name}) has already been imported from"
+                            Autoproj.warn "        #{already_loaded_vcs}"
+                            Autoproj.warn "        with server configuration '#{repository_configuration_label(already_loaded_vcs)}'."
```
doudou: I don't get that one. Isn't the VCS object enough ?

lib/autoproj/ops/configuration.rb
```diff
@@ -260,19 +290,23 @@ def load_and_update_package_sets(root_pkg_set, only_local = false)
...
+        def cleanup_remotes_dir(package_sets = manifest.package_sets, required_remotes_dirs = Array.new)
             # Cleanup the .remotes and remotes_symlinks_dir directories
             Dir.glob(File.join(remotes_dir, '*')).each do |dir|
                 dir = File.expand_path(dir)
-                if File.directory?(dir) && !package_sets.find { |pkg| pkg.raw_local_dir == dir }
-                    FileUtils.rm_rf dir
+                # Once a package set has been checked out during the process,
+                # keep it -- so that it won't be checked out again
+                if !required_remotes_dirs.include?(dir)
+                    if File.directory?(dir) && !package_sets.find { |pkg| pkg.raw_local_dir == dir }
```
doudou: The package sets test is redundant here, remove it.

lib/autoproj/ops/configuration.rb
```diff
@@ -260,19 +290,23 @@ def load_and_update_package_sets(root_pkg_set, only_local = false)
                 queue_auto_imports_if_needed(queue, pkg_set, root_pkg_set)
             end
 
-            cleanup_remotes_dir(package_sets)
+            cleanup_remotes_dir(package_sets, required_remotes_dirs)
             cleanup_remotes_user_dir(package_sets)
```
doudou: I would give the same treatment to cleanup_remotes_user_dir

lib/autoproj/ops/configuration.rb
2maz: I am trying to make the error message readable to a user!
Since the user only does:
 -- Autoproj.gitorious_server_configuration('GITORIOUS', 'gitorious.org')
Rendering it into a url e.g. repository_id=gitorious://... is an autoproj interna.

lib/autoproj/ops/configuration.rb
2maz: Why would that be necessary -- if I understand it right it already cleans up all dead symlinks

doudou: My problem with it is that it goes the long route (implicitely through the package sets) to determine what are irrelevant symlinks when there is an easy way to go the short route (add it to a set whenever call the symlink update method)

lib/autoproj/ops/configuration.rb
doudou: The actual string that the user wrote, as well as the list of overrides that got applied to it, is already stored in the vcs object (under VCSDefinition#raw). You can have a look here (https://github.com/rock-core/autoproj/blob/autoproj-2.0/lib/autoproj/cli/show.rb) to see how it is being interpreted by autoproj show.

2maz: The vcs object is used here already. Just going through options[:repository id] which looks to me like a more stable interface in the given case!

``` ruby
[1] pry(#<Autoproj::Ops::Configuration>)> p vcs
#<Autoproj::VCSDefinition:0x00000001bb8838 @options={:push_to=>"git@github.com:/rock-core/package_set.git", :retry_count=>10, :repository_id=>"github:/rock-core/package_set.git"}, @url="git://github.com/rock-core/package_set.git", @type="git", @raw=[[nil, {"github"=>"rock-core/package_set"}]]>
=> #<Autoproj::VCSDefinition:0x00000001bb8838
 @options={:push_to=>"git@github.com:/rock-core/package_set.git", :retry_count=>10, :repository_id=>"github:/rock-core/package_set.git"},
 @raw=[[nil, {"github"=>"rock-core/package_set"}]],
 @type="git",
 @url="git://github.com/rock-core/package_set.git">
[2] pry(#<Autoproj::Ops::Configuration>)>

doudou: If what you want is not confuse the user, showing him what he exactly wrote is by far the best option. If anything, repository_id is the internal parameter (used internally to compare if two VCS are from the same repository).

2maz: I can remove the server configuration information, but I would not want to invest much more time on that topic!? 

Current message see below. 
``` ruby
  WARN: redundant auto-import by main configuration for package set 'rock.dfki'.
  WARN:     A package set with the same name (rock.dfki) has already been imported from
  WARN:         git:git@git.hb.dfki.de:/rock-dfki/package_set.git push_to=git@git.hb.dfki.de:/rock-dfki/package_set.git repository_id=dfkigit:/rock-dfki/package_set.git retry_count=10
  WARN:         with server configuration 'dfkigit'.
  WARN:     Skipping the following one: 
  WARN:         git:git@git.hb.dfki.de:/rock-dfki/package_set.git push_to=git@git.hb.dfki.de:/rock-dfki/package_set.git repository_id=spacegit:/rock-dfki/package_set.git retry_count=10
  WARN:         with server configuration 'spacegit'.

doudou: > I can remove the server configuration information, but I would not want to invest much more time on that topic!?

Then remove. I think that "with server configuration" message is as confusing as no message at all. The only really useful thing for the user would be the file and line number where it happens, or at least the file and exact string to look for. Which is the information in 'raw' is.

2maz:
removed the "with server configuration" part

PR #90, opened 2015-09-15
support for msys2 installations on win64

planthaber:
"msys2" (https://msys2.github.io/) is a build environment for mingw on windows that includes a package manager (based on cygwin and mingw)

### DIFF
--- a/lib/autoproj/system.rb
+++ b/lib/autoproj/system.rb
@@ -6,6 +6,13 @@ module Autoproj
     def self.create_symlink(from, to)
         if Autobuild.windows?
             Dir.create_junction(to, from)
+        elsif Autobuild.msys?
+            begin
+                #looks like force option does not work
+                FileUtils.ln_sf from, to
+            rescue Errno::EEXIST
+                Autobuild.message "link exists, ignorig"
+            end
         else
             FileUtils.ln_sf from, to
         end
--- a/bin/autoproj_bootstrap
+++ b/bin/autoproj_bootstrap
@@ -16,6 +16,11 @@ module Autobuild
     def self.macos?
         @macos
     end
+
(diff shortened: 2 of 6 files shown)


lib/autoproj/system.rb
```diff
@@ -6,6 +6,14 @@ class UserError < RuntimeError; end
...
         if Autobuild.windows?
             Dir.create_junction(to, from)
+        elsif Autobuild.msys?
+            begin
+                #msys created a copy instead of a link
+                FileUtils.ln_sf from, to
+            rescue Errno::EEXIST
+                FileUtils.rm_rf to
+                FileUtils.ln_sf from, to
+            end
```
planthaber: I don't really know what to do here, msys2 internally avoids symlinks and is doing a copy instead, so at the second execution, FileUtils.ln_sf throws Errno::EEXIST.

Ignoring the call leads to non-updating autoproj/remotes folders, deleting them works, but local changes could be lost

goldhoorn: What if you use the windows way?

             Dir.create_junction(to, from)

planthaber: This function is available in the win32-dir gem, but sadly, that is not compatible with msys2.

goldhoorn:
@doudou: do you want a rebase against autobuild2.0 otherwise :+1: for merge

doudou:
> @doudou: do you want a rebase against autobuild2.0

Against autoproj-2.0, yes. I'm not going to release any new 1.x release (well, probably one, but with zero new functionality, just avoiding auto-update to autoproj 2.0)

goldhoorn:
@planthaber: can you update this PR?

planthaber:
It is on the agenda, but Autoproj 2 moved a lot of changed functions to separate files which makes merging more difficult as i have to find the new locations first

PR #120, opened 2016-09-22
Return cli status code

g-arjones:
This is important for automating bootstrap tests

### DIFF
--- a/bin/autoproj_bootstrap
+++ b/bin/autoproj_bootstrap
@@ -601,4 +601,4 @@ if !ops.skip_stage2?
     ops.call_stage2
 end
 ops.run_autoproj 'bootstrap', *bootstrap_options
-
+exit $?.exitstatus
--- a/bin/autoproj_bootstrap.in
+++ b/bin/autoproj_bootstrap.in
@@ -19,4 +19,4 @@ if !ops.skip_stage2?
     ops.call_stage2
 end
 ops.run_autoproj 'bootstrap', *bootstrap_options
-
+exit $?.exitstatus


bin/autoproj_bootstrap
```diff
@@ -601,4 +601,4 @@ if !ops.skip_stage2?
     ops.call_stage2
 end
 ops.run_autoproj 'bootstrap', *bootstrap_options
```
doudou: I would prefer that you change `#run_autoproj` to raise if the call fails (`Kernel#system` returns true/false on success/failure)

g-arjones: Couldn't the exit status be useful?

doudou: Not as far as I can see. Autoproj's really exiting either with 0 or 1

g-arjones: Ok. So raise in `#run_autoproj` and rescue somewhere? Btw, I would have to modify both autoproj_bootstrap and ops/install.rb, right?

g-arjones: No concerns about other methods using run_autoproj or backward compatibility issues?

doudou: > and rescue somewhere?

Not sure it's really useful to rescue in this case, since you will get your non-zero exit status.

> No concerns about other methods using run_autoproj or backward compatibility issues?

v2 is still in RC, and any method that fails should raise (i.e. current behaviour is unexpected).

g-arjones: > Not sure it's really useful to rescue in this case, since you will get your non-zero exit status.

... and a not very descriptive uncaught exception error message. Is that desirable in a script?

PR #121, opened 2016-09-26
Autodetect ruby version only if it is not specified

2maz:
To perform tests or computing dependencies for a given ruby version
'autodetect_ruby' should not override previous settings.

### DIFF
--- a/lib/autoproj/osdeps.rb
+++ b/lib/autoproj/osdeps.rb
@@ -950,10 +950,12 @@ fi
         end
 
         def self.autodetect_ruby
-            self.alias(ruby_version_keyword, "ruby")
+            if !aliases.has_key?("ruby")
+                self.alias(ruby_version_keyword, "ruby")
+            end
         end
-	self.suffixes << ruby_version_keyword
         autodetect_ruby
+        self.suffixes << aliases["ruby"]
 
         AUTOPROJ_OSDEPS = File.join(File.expand_path(File.dirname(__FILE__)), 'default.osdeps')
         def self.load_default


doudou:
This would be fixing the symptom, and not the problem. The problem does not lie in the alias creation, but in `#ruby_version_keyword` which should really not change during autoproj's execution.

2maz:
So you would prefer just setting ruby_version_keyword before initialization of Autoproj?

doudou:
I'm not sure what you're really trying to achieve, actually ...

2maz:
The actual use case here is to identify the required osdeps for other platforms than the current one, e.g. running 14.04 but generating the dependency list for 16.04.
While that is easily done with autoproj for the osdeps in general (setting Autopoj::OSDependencies.operating_system), the ruby version also needs to match for the target platform and the current implementation of ruby autodetection prevents a custom setting.

doudou:
OK ... got it.

First of all, I have a strong dislike with any kind of changes on the v1 branch. It's from my point of view a "the slowest maintenance mode possible", and I don't want to put any changes on it that could break something else - given how fragile v1 is, that means any change really.

In other words: please consider using v2 for non-standard workflows. If I were you, I would also consider installing more recent ruby versions on your older 14.04 installs using e.g. rbenv... 2.1+ generational GC *is* a major improvement.

That being said, I wonder if you could achieve exactly the same result by manually overriding the alias and the suffixes in init.rb with v1. I'm refactoring these bits in v2 to make it easier to achieve the same result.

doudou:
Closing as "too old to be worth dev time" now that v2 has been released for a while. Re-open if you don't manage to achieve what you want on v2.

PR #142, opened 2017-07-14
Change the link in the README.md for the autoproj_bootstrap v.1.X

Rauldg:
should point to the v1 branch instead of to the 1.13.3 tag


doudou:
Forgot that it was linked from there. Thanks ! :+1:

PR #155, opened 2017-12-04
avoid leaking backtraces to users, for errors that are most likely caused by them

doudou:
All CLI errors that are not internal errors are now subclasses of
CLIException (the same way that Autobuild errors are subclasses of
ConfigError). This ensures that we can rescue the error and show it
without a backtrace at the root.


g-arjones:
👍 

I am using this in my main autoproj bootstrap. Will let you know if I catch more

PR #158, opened 2017-12-08
fix type error leading to error message: package_managers/shell_scrip…

vbargsten:
…t_manager.rb:118:in `generate_auto_os_script': undefined method `join' for "zypper --non-interactive install --type pattern '%s'":String (NoMethodError)

Not 100% sure if the type is now as intended. Other package managers should be checked for this problem, too.


doudou:
Indeed, the yum importer had the same problem

PR #162, opened 2018-01-11
cli: make sure that 'exec' does not eat the subcommand's option

doudou:


g-arjones:
This solves the issue 👍

PR #163, opened 2018-01-11
cli: fix environment used in 'exec'

doudou:


g-arjones:
How was this corrupting env.sh as well?

doudou:
> How was this corrupting env.sh as well?

Wasn't. The workspace was not completely built, and I interpreted it as coming from the same bug.

g-arjones:
Got it.. 👍 then

PR #168, opened 2018-02-06
envsh: add option to watch workspace changes

g-arjones:
Pending issues:

 - [x] fix `#update_workspace`
 - [x] add tests

@doudou Thoughts? How do I fix `#update_workspace`?

### DIFF
--- a/lib/autoproj/cli/envsh.rb
+++ b/lib/autoproj/cli/envsh.rb
@@ -1,19 +1,151 @@
+require 'rbconfig'
 require 'autoproj/cli/inspection_tool'
 module Autoproj
     module CLI
         class Envsh < InspectionTool
-            def validate_options(_unused, options = Hash.new)
-                _, options = super(_unused, options)
+            attr_predicate :watch?
+            attr_reader :options
+            attr_reader :source_packages
+            attr_reader :notifier
+            attr_reader :watchers
+            def validate_options(unused, options = {})
+                _, options = super(unused, options)
+                @watchers = []
+                @watch = options[:watch]
+                @options = options
                 options
             end
 
-            def run(**options)
+            def update_workspace(create = false)
+                if create
+                    new_ws = Autoproj::Workspace.new(ws.root_dir)
+                    new_ws.set_as_main_workspace
+                    @ws = new_ws
+                end
                 initialize_and_load
-                shell_helpers = options.fetch(:shell_helpers, ws.config.shell_helpers?)
-                finalize_setup(Array.new)
+                shell_helpers = options.fetch(:shell_helpers,
+                                              ws.config.shell_helpers?)
+                @source_packages, = finalize_setup([])
                 export_env_sh(shell_helpers: shell_helpers)
+            end
+
+            def callback
+                puts 'Workspace changed...'
+                stop_watchers
+                update_workspace(true)
+                start_watchers
+            end
+
+            def create_file_watcher(file)
+                watchers << notifier.watch(file, :modify) do |e|
+                    delete_watcher(e.watcher) if e.flags.include? :ignored
+                    callback
+                end
+            end
+
+            def create_dir_watcher(dir, *files, recursive: false)
+                opt_args = %i[move create delete]
+                opt_args << :recursive if recursive
+
+                watchers << notifier.watch(dir, *opt_args) do |e|
+                    delete_watcher(e.watcher) if e.flags.include? :ignored
+                    file_name = File.basename(e.absolute_name)
+                    next unless files.any? { |regex| file_name =~ regex }
+                    callback
+                end
+            end
+
+            def delete_watcher(watcher)
+                watchers.delete(watcher)
+            end
+
+            def create_src_pkg_watchers
+                source_packages.each do |pkg_name|
+                    pkg = ws.manifest.find_autobuild_package(pkg_name)
+                    manifest_file = File.join(pkg.srcdir, 'manifest.xml')
+
+                    next unless File.exist? pkg.srcdir
+                    create_dir_watcher(pkg.srcdir, /^manifest.xml$/)
+                    next unless File.exist? manifest_file
+                    create_file_watcher(manifest_file)
+                end
+            end
+
+            def create_autobuild_watchers(dir)
+                Dir[File.join(dir, '*.autobuild')].each do |file|
+                    create_file_watcher(file)
+                end
+            end
+
+            def create_manifests_watchers(dir)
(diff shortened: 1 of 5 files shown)


doudou:
First, given how little similarity there is between envsh and `envsh --watch`, I'd go for creating a new subcommand (possibly "advanced" so that it doesn't show up in the --help).

I'll have a look about `update_workspace`. The immediate problem I see is with the way autobuild uses Rake (which I should have fixed a long time ago) and with the global autobuild package registry (which should go away). Is there anything else ?

The quick & dirty way to avoid these issues would be to have autoproj exec itself when anything changes...

g-arjones:
> I'd go for creating a new subcommand

Sounds good...

> with the global autobuild package registry

That's what's breaking the method...

> Is there anything else ?

Not that I know of.

> The quick & dirty way to avoid these issues would be to have autoproj exec itself when anything changes...

We still need an updated instsance of `Autoproj::Workspace` so we can update the watchers in case new packages/package sets get added..

g-arjones:
> One thing, though, you would need to set up the watchers before the initial update, or there's a window where you would miss an update.

With `exec` there is no way to prevent that. Events triggered while the process is being replaced and before the watchers are setup _will_ be missed, regardless of how early the setting up happens. That said, the earliest, the better.

Will update.

doudou:
> With exec there is no way to prevent that. Events triggered while the process is being replaced and before the watchers are setup will be missed, regardless of how early the setting up happens. That said, the earliest, the better.

Since you do an initial update in `watch`, this is not an issue:

$ autoproj watch
1. Setup watch
2. Initial update
3. Wait
4. Got event -> Exec
5. Setup watch
6. Initial update <- this update handles the event that triggered the exec *and* any event that might have happened between 4 and 5

g-arjones:
> this update handles the event that triggered the exec and any event that might have happened between 4 and 5

Good point! 👍

lib/autoproj/cli/envsh.rb
```diff
@@ -1,19 +1,147 @@
...
+                _, options = super(unused, options)
+                @watchers = []
+                @watch = options[:watch]
+                @options = options
                 options
             end
 
-            def run(**options)
+            def update_workspace(restart = false)
+                exec('autoproj', 'envsh', '--watch') if restart
```
doudou: This method is now doing two completely unrelated things based on the restart flag. Separate the methods into `update_workspace` and `restart` (and *definitely* split envsh from watch).

lib/autoproj/cli/envsh.rb
```diff
@@ -1,19 +1,147 @@
...
+
+            def create_file_watcher(file)
+                watchers << notifier.watch(file, :modify) do |e|
+                    delete_watcher(e.watcher) if e.flags.include? :ignored
+                    callback
+                end
+            end
+
+            def create_dir_watcher(dir, *files, recursive: false)
+                opt_args = %i[move create delete]
```
doudou: I would prefer having the static arguments directly in the call for clarity:

`watchers << notify.watch(dir, :move, :create, :delete, *opt_args) ...`

g-arjones:
I have to load the workspace before setting up the watchers otherwise I won't know what to watch :(

doudou:
> I have to load the workspace before setting up the watchers otherwise I won't know what to watch :(

Ah well ... Let's assume nobody will manage to save just at the wrong time.

For what it's worth, I noticed something in Listen that might bite us here (I'm mentioning it so that we keep it in mind): they have a fair bit of logic dedicated to transforming an editor's atomic save (which is basically a write -> rename) into a single modify event. I'm guessing that this logic should handle it fine (it would trigger on rename), but I'm not entirely sure right now.

g-arjones:
> I noticed something in Listen

Not using Listen anymore..

We have a dir watcher for every file watcher so I don't see how that could be a problem..

doudou:
> Not using Listen anymore..

Which is why I'm mentioning it. I wouldn't mention it otherwise, as the complexity (fs-event-wise) of a simple file save is not hidden from us. A simple file save can lead to many events, and I want to write it down so that this is documented.

Simple example - which would not be so far-fetched if we weren't exec'ing - is that we would try to load the new workspace at a point where the file has been moved but not yet replaced, which would be invalid. Wouldn't happen often, but would happen.

A.k.a. there's always a cost to be close to the metal. You don't benefit from years of knowledge from people that are doing *exactly* what we're tying to do. Listen's main use-case is guard, which is basically a "run something when something changes" environment.

doudou:
Looks good .. only missing tests now

g-arjones:
Great.. I got caught by some "urgent" (?) stuff and could not finish them but they are super easy to write, autoproj's test suite already has all helpers I need I just did not have time to.. Will probably have to wait until after carnival now :P

g-arjones:
Updated.

doudou:
*Really* cool. Thanks a lot. I'd like to test it out here before I merge, but it looks good.

g-arjones:
Sounds good. Thanks..

doudou:
Houston, we have a problem ...

A lot of autoproj commands (namely, autoproj exec) are saving the config file which triggers watch ... and would probably trigger envsh which would save the config file :P. I'm looking into that.

doudou:
.... and I already got concurrent file modification on `config.yml` ...

g-arjones:
Damn.. Did not expect that. Thanks for having a look

g-arjones:
> .... and I already got concurrent file modification on config.yml ...

That is something I would expect but seemed unlikely. How did you do that? We probably need a file lock

doudou:
> That is something I would expect but seemed unlikely. How did you do that? We probably need a file lock

Since all commands write config.yml, it's a lot easier than it sounds. I had `autoproj watch` and Ruby linters running under `autoproj exec`

I'm working on all of this (and adding tests for Configuration ... this is really my cross to bear). Could you please try to create branches and pull requests directly on the main repositories ? It would make things a lot easier as I could push directly to the PR instead of having to create a new one.

I'm moving to use atomic writes instead of file locks for the time being

g-arjones:
> Could you please try to create branches and pull requests directly on the main repositories 

Don't have write permission :(

doudou:
> Don't have write permission :(

Then ask for it... I thought you did.

doudou:
The "rock core" team somehow did not have write permissions on autoproj ... Just fixed it.

g-arjones:
Would you prefer to close this and open another one?

doudou:
> Would you prefer to close this and open another one?

Would be great. Just refer to this thread in the PR message.

PR #169, opened 2018-02-08
README: Better readability for installation instructions when checking out autoproj manually

maltewi:
Better readability for installation instructions when checking out autoproj manually


doudou:
Thanks !

PR #170, opened 2018-02-16
Watch workspace changes

g-arjones:
Follow-up from: #168


doudou:
I've pushed some updates that fix the atomicity for me, and avoid unnecessary config file events

g-arjones:
Great! Thanks..

PR #171, opened 2018-02-16
Fix packages versions management

g-arjones:


lib/autoproj/autobuild_extensions/git.rb
```diff
@@ -87,7 +87,7 @@ def snapshot_against_remote(package, options = Hash.new)
                 if options[:exact_state] && !info['tag']
                     info['commit'] = rev_parse(package, 'HEAD')
                 end
-                info
+                info.delete_if { |_, v| v.nil? }
```
doudou: The `nil`'s here as important. Without them, the override would not remove e.g. a tag that would have been set in a package set.

lib/autoproj/cli/commit.rb
```diff
@@ -5,34 +5,48 @@
...
                 if !importer || !importer.kind_of?(Autobuild::Git)
-                    raise CLIInvalidArguments, "cannot use autoproj tag if the main configuration is not managed by git"
+                    raise CLIInvalidArguments, "cannot use autoproj commit if the main configuration is not managed by git"
                 end
 
-                versions_file = File.join(
-                    ws.config_dir,
-                    Workspace::OVERRIDES_DIR,
-                    Versions::DEFAULT_VERSIONS_FILE_BASENAME)
+                unless tag_name.nil?
```
doudou: This is `if !tag_name.nil?`. A.k.a. `if tag_name`.

Two style constraints in packages I maintain:
- don't use `nil?` unless you really mean it. In this case, the "thruthiness" of tag_name is enough
- don't use unless, ever.

g-arjones: > don't use nil? unless you really mean it

I try my best to follow your style when I am working on stuff that are mostly maintained by you alone. This can be really frustrating sometimes.. In this particular case I did it like this because you did it like this yourself [here](https://github.com/rock-core/autoproj/blob/1fdb3230869325f246ec110ff7f24647c530dbbc/lib/autoproj/cli/tag.rb#L24)

> don't use unless, ever.

This one was really an attempt to follow a community-adopted common practice (suggested by Rubocop) but I will remove it.

lib/autoproj/autobuild_extensions/git.rb
g-arjones: Doesn't the `commit` field do that?

lib/autoproj/cli/commit.rb
doudou: > I try my best to follow your style when I am working on stuff that are mostly maintained by you alone. This can be really frustrating sometimes..

I honestly don't know why there is a `nil?` here. I never do this. Maybe refactoring of code from other people ... Sorry for that one.

> This one was really an attempt to follow a community-adopted common practice (suggested by Rubocop) but I will remove it.

I'm tuning the Rubocop rules to better fit the style in my packages. There are quite a bit of Rubocop rules that are actually not "accepted best practices".

I've been using `unless` for a while, but the logic gets hard to follow ... one ends up with double-negations quite often (like here)

lib/autoproj/autobuild_extensions/git.rb
doudou: > Doesn't the commit field do that?

But, then, how do you remove the commit to favor a tag ?

lib/autoproj/cli/commit.rb
doudou: OK, let's soften this a bit: ultimately, this is the style autoproj is coded with (mostly) because that's the one I prefer. In the spirit of being welcoming (here, mostly for you), I would not mind adopting the default rubocop ruleset (minus some stuff that is even listed as "both ways are widely accepted in the ruby style guide"). A.k.a. I'm open for discussion here.

Setting up and using rubocop would definitely help with consistency, on both sides. I do understand that it's frustrating to have to guess how the style should be...

g-arjones: > I've been using unless for a while, but the logic gets hard to follow ... one ends up with double-negations quite often (like here)

I only favor `unless` over `if not` and I think (can't say for sure) that's what's suggested in ruby's style guide.

> using rubocop would definitely help with consistency, on both sides

👍  ... I've been doing this for a while I the only thing I override is the indentation width (two spaces really ruins readability for me)

> I do understand that it's frustrating to have to guess how the style should be

I'm glad you got exactly what I meant. I have no problems with following someone else's style in situations like these, the problem is when as you very well said it basically becomes guess work.

lib/autoproj/autobuild_extensions/git.rb
g-arjones: Not sure I understand. The generated version file will contain `commit` or `tag` (never both). So, if autoproj really favors what is set in the override file, that should remove whatever is set in the package_set. Am I missing something?

lib/autoproj/cli/commit.rb
g-arjones: > There are quite a bit of Rubocop rules that are actually not "accepted best practices".

We should probably open issues in Rubocop's repo for those cases.

```
RuboCop is a Ruby static code analyzer. Out of the box it will enforce many of the guidelinesoutlined in the community Ruby Style Guide.
```

doudou: > I've been doing this for a while I the only thing I override is the indentation width (two spaces really ruins readability for me)

I started doing it the other day, but then vscode became horribly sluggish. There's a PR on vscode-ruby to rate-limit the calls to rubocop, but it's not merged/released yet.

doudou: > > There are quite a bit of Rubocop rules that are actually not "accepted best practices".

> We should probably open issues in Rubocop's repo for those cases.

Regardless of what the rubocop guys say, things like the `if !` vs unless is more split than they acknowledge.

There are quite a few rules where they say themselves than any settings for a rule are widely accepted.

lib/autoproj/autobuild_extensions/git.rb
doudou: > Am I missing something?

g-arjones: Perfectly clear. Thanks!

doudou: For reference: https://doudou.github.io/rock_website/workspace/add_packages.html#version_control_resolution

g-arjones:
Updated.

lib/autoproj/cli/main.rb
```diff
@@ -343,12 +343,12 @@ def tag(tag_name = nil, *packages)
...
             end
 
-            desc 'commit [PACKAGES]', 'save the package current versions as a new commit in the main build configuration'
+            desc 'commit [TAG_NAME] [PACKAGES]', 'save the package current versions as a new commit in the main build configuration'
             long_desc <<-EOD
 The commit subcommand stores the state of all packages (or of the packages
 selected on the command line) into a new commit in the currently checked-out
 branch of the build configuration. This state can be retrieved later on by using
-"autoproj reset"
+"autoproj reset". If a TAG_NAME is provided, the commit will be tagged.
```
doudou: Sorry, I just realized this.

You cannot provide a package without a tag anymore, since the first argument is always interpreted as a tag.

You would either need to reintroduce the `tag` subcommand, or add a `--tag` option to `commit`

g-arjones: The tag subcommand has the same issue. I assumed it was a known limitation. Will fix both

g-arjones: Nevermind.. [TAG_NAME] is not really optional for the tag subcommand.

g-arjones: Done.

doudou: Wait ... if there is a `tag` subcommand ... why are you introducing the ability to create a tag in `commit` ? Or do you think it's better this way and that we should deprecate `tag` ? I'm not sure myself, as we still need a command to list the tags ...

g-arjones: Maybe merge both.. Currently, `tag` doesn't commit and `commit` doesn't tag :P

g-arjones: `tag` just creates a dangling commit...

g-arjones: Or, deprecate `tag`. I don't see a good use case for that even though they do different things...

doudou: Alright, now I see it. Thanks for the refresher between `commit` and `tag`

`tag` does not create a dangling commit, it creates a `tag` WITHOUT changing the current branch. Which is great to tag "known good states" without pinning the branch - i.e. letting everyone continue working normally. 

`commit` publishes a certain state to the branch. In the time where I was still using those, `tag` was actually a lot more useful for me than `commit.

I do agree that having the new `--tag` option to `commit` is useful, so I'm still fine with this PR. But the two commands are still very different.

g-arjones: With `tag` the generated version file is still commited just not to any branch (otherwise tagging wouldn't be possible). That what I meant by a dangling commit.

> I do agree that having the new --tag option to commit is useful, so I'm still fine with this PR. But the two commands are still very different.

How would you like to proceed with this?

doudou: > That what I meant by a dangling commit.

Got that ... But then sorry for being pedantic, but "dangling commit" in Git really means "not attached to a branch or tag". The generated commit for instance won't show up in a `git fsck`.

doudou:
> How would you like to proceed with this?

Well ... by merging the PR I guess ;-)

g-arjones:
Thought as much.. Thanks :)

PR #172, opened 2018-02-16
cli: do not abort 'status' on remote errors when --snapshot is set

doudou:
WHile the main import was ignoring all exceptions, the snapshotting
was ignoring Autobuild::PackageException but not handling the rest.
It ignores these because there are a lot of cases where snapshotting
is not possible (i.e. if the commit is not present remotely).

Keep this case, but warn the user for all error cases instead of
aborting with an internal error


g-arjones:
👍

PR #177, opened 2018-02-26
update dependencies

doudou:
Depends on:
- [ ] https://github.com/rock-core/autobuild/pull/44

When running a dev setup (that is, a setup where autoproj and autobuild are referring
to local folders), one MUST:
- update both autoproj and autobuild
- **without loading the env.sh**:
  - run .autoproj/bin/bundle update


g-arjones:
:+1:

PR #178, opened 2018-02-26
fix configuration not being dirtied on modification to utilities

doudou:
The symptom was that `autoproj test enable ...` or
`autoproj test disable ...` would basically do nothing.

### DIFF
--- a/autoproj.gemspec
+++ b/autoproj.gemspec
@@ -26,6 +26,7 @@ Gem::Specification.new do |s|
 
     s.add_runtime_dependency "bundler"
     s.add_runtime_dependency "autobuild", ">= 1.12.0"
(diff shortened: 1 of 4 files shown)


g-arjones:
Good... I wonder when did this brake..

doudou:
> I wonder when did this brake..

When I added the "save configuration only when modified" thingy. Actually, a good way to ensure it never happens again would be to `dup` the value returned by `get`. Thoughts ?

g-arjones:
> would be to dup the value returned by get

That's better, IMO. We don't want anyone updating this hash without going through the API anyway...

doudou:
Updated

g-arjones:
Looks good... Live tested.

PR #180, opened 2018-02-26
cli: report which packages failed after a build failure under --tool

doudou:
The output is really confusing, and it really helps to have
a summary of which packages failed at the end - especially in tools like e.g. vscode that
sometimes do not have a reporter for some errors.

### DIFF
--- a/lib/autoproj/cli/main.rb
+++ b/lib/autoproj/cli/main.rb
@@ -35,13 +35,13 @@ module Autoproj
                     end
                 end
 
-                def run_autoproj_cli(filename, classname, report_options, *args, **extra_options)
+                def run_autoproj_cli(filename, classname, report_options, *args, tool_failure_mode: :exit_silent, **extra_options)
                     require "autoproj/cli/#{filename}"
                     if Autobuild::Subprocess.transparent_mode = options[:tool]
                         Autobuild.silent = true
                         Autobuild.color = false
                         report_options[:silent] = true
-                        report_options[:on_package_failures] = :exit_silent
+                        report_options[:on_package_failures] = tool_failure_mode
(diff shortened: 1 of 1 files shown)


g-arjones:
This improves the situation a bit but when multiple packages are being built in parallel the output is still pretty bad as the processes are racing each other for STDIO. What do you think about buffering the output until the process ends and synchronizing the threads so that messages do not get interleaved?

doudou:
> What do you think about buffering the output until the process ends and synchronizing the threads so that messages do not get interleaved?

I thought about it.

The downside is that, for package that take a long time to build, one gets the problems listed only at the end. Given how horrible the output currently is, it might be a small price to pay.

Ultimately, one shouldn't have to browse this output, though. So I'm not sure where would be the best use of our time (improve the matchers vs. synchronize the output)

g-arjones:
> one gets the problems listed only at the end

I don't mind that at all. What may be annoying is not getting any cue that something is going on in the background.

> improve the matchers vs. synchronize the output

Definitely, improve the matchers. I just made the point because it made sense since you decided to improve the output. But this is not bothering me at all so I guess it is up to you. What this PR does is already a step in the right direction..

doudou:
> I just made the point because it made sense since you decided to improve the output.

Basically how I do it with this patch, if the problem matchers don't work, is to trigger the build only of the failing package.

doudou:
> What may be annoying is not getting any cue that something is going on in the background.

Good point, we would need to implement a spinner like for `--status`. Definitely not trivial under the current architecture.

g-arjones:
Ok then, I suggest we merge this and keep on extending the problem matchers as we see the need until one of us get the time to improve `autoproj --tool` further..

PR #182, opened 2018-03-16
fix modified? being reset when writing an existing value

doudou:
As the saying goes, there are only two difficult things in
software engineering: naming things and cache invalidation.

https://martinfowler.com/bliki/TwoHardThings.html


g-arjones:
:smile:

PR #183, opened 2018-03-22
Add ros manifest support

g-arjones:


lib/autoproj/package_manifest.rb
```diff
@@ -25,11 +26,16 @@ def self.load(package, file)
         # @param [PackageDescription] the package we're loading it for
         # @param [String] contents the manifest.xml contents as a string
         # @param [String] path the file path, used for error reporting
+        # @param [Boolean] ros_manifest whether the file follows the ROS format
         # @return [PackageManifest]
         # @see load
-        def self.parse(package, contents, path: '<loaded from string>')
+        def self.parse(package, contents, path: '<loaded from string>', ros_manifest: false)
```
doudou: I would prefer that you pass the manifest loader, either directly as a class, or as a factory function (the latter would be probably better since it does not force the loader to have the same `#initialize` signature).

lib/autoproj/package_manifest.rb
```diff
@@ -216,5 +222,41 @@ def tag_end(name)
...
+                elsif name == 'tags'
+                    @tag_text = ''
+                else
+                    @tag_text = nil
+                end
+            end
+
+            def tag_end(name)
+                if DEPEND_TAGS.include?(name)
+                    manifest.add_dependency(@tag_text)
```
doudou: Autoproj does support `test_depend`

lib/autoproj/package_manifest.rb
```diff
@@ -216,5 +222,41 @@ def tag_end(name)
...
+                elsif AUTHOR_FIELDS.include?(name)
+                    manifest.send("#{name}s").concat(parse_contact_field(@tag_text))
```
doudou: package.xml's author field is formatted differently from manifest.xml's (the email is stored as attribute in `package.xml`)

test/test_manifest.rb
```diff
@@ -684,6 +690,68 @@ module Autoproj
                 end
                 assert_equal "manifest #{File.join(ws.root_dir, 'test', 'manifest.xml')} of test from pkg_set lists 'dependency' as dependency, but it is neither a normal package nor an osdeps package. osdeps reports: cannot resolve dependency: dependency is not an osdep and it cannot be resolved as a source package", e.message
             end
+            it "parses an autoproj format manifest" do
```
doudou: General comments about the tests:
- I know you copied the way the other tests were doing it (mocking everything), which is obviously horrible. Given that the rest of the file already does it, I can't ask you to change it, but I would nowadays go for actually writing files to disk instead.
- you don't test the ROS loader itself, only the conditional logic that would attempt to load `package.xml`

lib/autoproj/package_manifest.rb
g-arjones: No problem, I can do that...

lib/autoproj/package_manifest.rb
g-arjones: Yeah, but that relies on the user actually using the `test` sub command. Given that (I am assuming) tests are layed out differently in catkin packages, I don't think that would be a good thing.

lib/autoproj/package_manifest.rb
g-arjones: Yes, but I autoproj doesn't use/store the email anywhere, does it? That's why I made the conscious decision of ignoring the email. Thoughts?

doudou: > Yes, but I autoproj doesn't use/store the email anywhere, does it? That's why I made the conscious decision of ignoring the email. Thoughts?

Autoproj does not. Tools layered on top of it does, but then in this case it will simply show no emails (it is supposed to be able to handle that).

lib/autoproj/package_manifest.rb
doudou: > Given that (I am assuming) tests are layed out differently in catkin packages, I don't think that would be a good thing.

Can't the ROS people at BIR help you with this ? I have the feeling that building tests and the rest of the package(s) is separated in catkin anyways.

test/test_manifest.rb
g-arjones: > which is obviously horrible

The "horrible" part is because these tests are screaming for a refactor, there are lots of repeated code.

> I would nowadays go for actually writing files to disk instead

Generally speaking, I prefer using mocks in unit tests to make them more "unittary", so to speak. This is part of the problem of orogen's/orocos.rb's tests, they are way too dependent on external stuff, turning them into a major pain.

The integration part is left to integration tests.

> you don't test the ROS loader itself, only the conditional logic that would attempt to load package.xml

I deemed that unnecessary since `RosLoader´ is just a stripped down version of the other `Loader` but I can write those as well. No problem

lib/autoproj/package_manifest.rb
g-arjones: I could also get the "email" attribute and join that in `@tag_text`. Would you prefer that?

lib/autoproj/package_manifest.rb
g-arjones: > Can't the ROS people at BIR help you with this ?

Not really, unfortunately.

> I have the feeling that building tests and the rest of the package(s) is separated in catkin anyways

Catkin macros' build them automatically if dependencies are met (unless you explicitly disable the tests). I don't there's a good way to make these macros and `autoproj test` play nicely together.

test/test_manifest.rb
doudou: > Generally speaking, I prefer using mocks in unit tests to make them more "unittary", so to speak. This is part of the problem of orogen's/orocos.rb's tests, they are way too dependent on external stuff, turning them into a major pain.

I was there, and came back. If the code needs to touch the file system, then let it touch the filesystem. This actually makes the test less, not more, brittle as a new implementation of the same logic will have to deal with the filesystem the same way.

I don't see value in having "pure unitary" unit tests (I believe that there is mostly no such thing in the real world), and believe that the unit/integration test dichotomy is mostly false. And I'm not alone in this. Mocking your internal interfaces is actually pointless, as it makes the tests horrible to follow because of the amount of mocking, and extremely brittle when refactoring.

The problem with orogen is that to test a code generator, the best thing you can do is to actually generate and compile the code - which takes *really* long. This is the reason why I started to create the rock-build-tests stuff - let amake do the building and then check the result. Leaving orogen to do unit tests on the processing (i.e. before the codegen)

For orocos.rb, there was no easy way to create tasks and then test against these tasks, which meant that I had to actually compile code and run subprocesses (again). Now that there are ruby task contexts, I would *definitely* not attempt to "improve" the tests by mocking, but by simply creating ruby task contexts to test against, which will make the tests a lot faster and a lot more self-contained.

Mocking only gives you an illusion of "isolation" of your tests, it exposes the internal interfaces in ways that sometimes make little sense. The one *hugely* profitable use of mocking for me is to reduce the number of branches I have to test against - when a method "resolves" a complex inputs into a normalized (smaller) set of outputs that are used by the method, I happily mock that to only have to deal with the smaller normalized set instead of repeating tests against the bigger argument set.

lib/autoproj/package_manifest.rb
doudou: > I could also get the "email" attribute and join that in @tag_text. Would you prefer that?

Wouldn't it make more sense to save it and then add the email directly (instead of using `parse_contact_field`) ?

g-arjones: Probably, yes. Will do that.

test/test_manifest.rb
g-arjones: > it makes the tests horrible to follow because of the amount of mocking, and extremely brittle when refactoring.

This became more of a problem with these highly dynamic languages that allow mocking of entire class definitions (and pretty much anything you want). People stopped being disciplined about dependency injection and factories. The reason most tests have to be rewritten when the implementation changes is because we are mocking internals even though we are testing interfaces. If we were injecting the dependencies, we would only have to rewrite the tests when the dependencies' interfaces changed (which is actually a good thing in these cases). It also has the good side effect of making you really feel that you are making a possibly API breaking change.

Of course there are pros and cons on both sides (just like static vs. dynamic typing). I used to regret these dynamic language but I started to appreciate them since Ruby.. TypeScript was also a good surprise in that regard.

doudou: > The reason most tests have to be rewritten when the implementation changes is because we are mocking internals even though we are testing interfaces. 

Agreed. Which is my original point. Test the interface ! It so happens that the filesystem is, for this class, part of the interface.

lib/autoproj/package_manifest.rb
doudou: (Questions below for my understanding, I'm totally OK with your argument - let's just handle all dependencies the same right now, and refine later)

> **automatically**

Yeah ... that sounds like catkin.

> (unless you explicitly disable the tests).

And you can't do that on a per-package basis ?

g-arjones: > And you can't do that on a per-package basis ?

I'm guessing there's probably a cmake definition that allows that but I didn't find it documented anywhere and given that ROS' documentation is quite good, if there's such API it is most likely not considered stable.

g-arjones:
Side question: any idea why autobuild keep passing `-DCMAKE_PREFIX_PATH=""` (empty) when configuring cmake packages even though I have `Autoproj.env_add_path 'CMAKE_PREFIX_PATH', '/opt/ros/kinetic'` in the build conf's init.rb?

If I do that in `cmake_package do` block it works as expected though...

doudou:
> Side question: any idea why autobuild keep passing -DCMAKE_PREFIX_PATH="" (empty) when configuring cmake packages even though I have Autoproj.env_add_path 'CMAKE_PREFIX_PATH', '/opt/ros/kinetic' in the build conf's init.rb?

This sounds like a potential bug ... `Autoproj.env_add_path` should be propagated.

g-arjones:
> This sounds like a potential bug ...

Could you please try to reproduce?

doudou:
> Could you please try to reproduce?

Sure ...

If I set the environment variable, globally or in the package, the values do not end up in the -D option. They do end up in the environment, though.

I just checked the code, and [the Autobuild::CMake class](https://github.com/rock-core/autobuild/blob/master/lib/autobuild/packages/cmake.rb) has its own `prefix_path` and `module_path` attributes (global and per-package). I'm not sure how CMake handles having paths provided by as environment and as definition, though.

doudou:
(One detail that might have derailed your tests: autobuild removes non-existent entries in path variables before it exports the env.sh)

g-arjones:
> (One detail that might have derailed your tests: autobuild removes non-existent entries in path variables before it exports the env.sh)

Not the problem as (1) the path exists and (2) the var _is_ ending up in the exported env.sh

> If I set the environment variable, globally or in the package, the values do not end up in the -D option. They do end up in the environment, though.

I'm confused.. How is this being handled in rock's package sets? We don't set CMAKE_MODULE_PATH and CMAKE_PREFIX_PATH for every package...

doudou:
> I'm confused.. How is this being handled in rock's package sets? We don't set CMAKE_MODULE_PATH and CMAKE_PREFIX_PATH for every package...

Good point. I guess that CMake uses both the `-D` and the environment variable at the same time, then.

Actually, just tested it out, and it does:

~~~
./prefix0
./prefix0/share
./prefix0/share/pkg0
./prefix0/share/pkg0/cmake
./prefix0/share/pkg0/cmake/pkg0-config.cmake
./prefix1
./prefix1/share
./prefix1/share/pkg1
./prefix1/share/pkg1/cmake
./prefix1/share/pkg1/cmake/pkg1-config.cmake
~~~

then the CMakeLists.txt

~~~
find_package(Pkg0)
find_package(Pkg1)
~~~

works with

~~~
CMAKE_PREFIX_PATH=$PWD/prefix0 cmake . -DCMAKE_PREFIX_PATH=$PWD/prefix1
~~~

As for the Rock packages, currently `base/cmake` globally sets the `CMAKE_PREFIX_PATH` environment variable to its prefix.

doudou:
(Oh, and the `prefix_path`/`module_path` accessors can be set for all cmake packages at the class level)

g-arjones:
> They're not, they are arrays, so you can do CMake.prefix_path << "value".

Yes, of course. Sorry. This works for me but I would prefer using `env_add_path` though. I have put on a test case just to make sure you get my point. Clone https://github.com/g-arjones/autoproj_env_test and do:

```sh
$ ruby autoproj_install 
$ source env.sh
$ aup --no-autoproj --local
$ amake dummy
```

Now:

```sh
head -3 install/log/dummy-configure.log
```

This gives:

```
2018-03-22 15:02:34 -0300: running
    cmake -DCMAKE_INSTALL_PREFIX=/home/arjones/env_test/install -DCMAKE_MODULE_PATH= -DCMAKE_PREFIX_PATH= /home/arjones/env_test/dummy
```
I expected `-DCMAKE_PREFIX_PATH=/home/arjones/env_test/install`

doudou:
> I expected -DCMAKE_PREFIX_PATH=/home/arjones/env_test/install

I get the confusion. Just to make sure that you got the gist of my point of view
- CMAKE_PREFIX_PATH can be provided both as an envvar and as a -D *at the same time*
- the envvar system is more powerful than `Autobuild::CMake`'s prefix_path

a.k.a. I don't see what problem there is left ... Just use `pkg.env_add_path` (and I need to add documentation about this, hence the new issue I created on the website)

g-arjones:
> I don't see what problem there is left 

Problem is that when one try to access `${CMAKE_PREFIX_PATH}` from within a cmake macro he/she gets an empty string rather than what is set in the environment. That obviously breaks things

doudou:
> Problem is that when one try to access ${CMAKE_PREFIX_PATH} from within a cmake macro he/she gets an empty string rather than what is set in the environment. That obviously breaks things

Except that the CMAKE_PREFIX_PATH environment variable is *not* initialized to the environment value, regardless of whether there is a -D argument passed. Just checked on my toy install, and also https://cmake.org/cmake/help/v3.0/variable/CMAKE_PREFIX_PATH.html.

A.k.a. from the p.o.v. of the CMake people, CMAKE_PREFIX_PATH should not even be set externally. Which is fine for us given that we both agreed that from Autoproj, using the environment should be the way to go.

g-arjones:
>  Which is fine for us given that we both agreed that from Autoproj, using the environment should be the way to go

Unfortunately, catkin guys don't agree with us :(
I will have to live with `Autobuild::CMake.prefix_path`

doudou:
> Unfortunately, catkin guys don't agree with us :(

Maybe define yourself a `catkin_package` helper that does the common setup (i.e. changes `pkg.prefix_path`) ?

Are you actually trying to build catkin packages but without using catkin itself ? You are much more brave than I ever was ...

g-arjones:
> Maybe define yourself a catkin_package helper that does the common setup (i.e. changes pkg.prefix_path) ?

That's a great idea. Will do that.

> Are you actually trying to build catkin packages but without using catkin itself ? You are much more brave than I ever was ...

Lol.. I guess I am. It's working surprisingly well actually. Most of the heavy lifting is done through cmake macros, the python module is basically a wrapper.

g-arjones:
@doudou Updated. Please let me know if you have further comments...

doudou:
Nope. All good.

g-arjones:
Cool... Would you mind releasing this, please?

doudou:
Not really, but ... why the rush (if I may ask) ?

g-arjones:
Makes it easier for other people to try it out here. No rush though, if you prefer to do your own testing that's ok..

PR #184, opened 2018-03-25
allow a source_dir configuration in the workspace

g-arjones:
The layout will be relative to `Autoproj.workspace.source_dir` rather than `Autoproj.workspace.root_dir` . If `Autoproj.workspace.config.source_dir` is unset then `source_dir == root_dir` (which is the default)

### DIFF
--- a/lib/autoproj/configuration.rb
+++ b/lib/autoproj/configuration.rb
@@ -381,6 +381,18 @@ module Autoproj
             get('build', 'build')
         end
 
+        # Defines a folder to which source packages will be layed out relative to
+        #
+        # If nil, packages will be layed out relative to root_dir
+        # Only relative paths are allowed
+        #
+        # The default is nil
+        #
+        # @return [String,nil]
+        def source_dir
+            get('source', nil)
+        end
+
         # Returns true if there should be one prefix per package
         #
         # The default is false (disabled)
--- a/lib/autoproj/ops/configuration.rb
+++ b/lib/autoproj/ops/configuration.rb
@@ -510,7 +510,7 @@ module Autoproj
             explicit.each do |pkg_or_set, layout_level|
                 next if manifest.find_autobuild_package(pkg_or_set)
                 next if manifest.has_package_set?(pkg_or_set)
-                full_path = File.expand_path(File.join(ws.root_dir, layout_level, pkg_or_set))
+                full_path = File.expand_path(File.join(ws.source_dir, layout_level, pkg_or_set))
                 next if !File.directory?(full_path)
 
                 if handler = auto_add_package(pkg_or_set, full_path)
--- a/lib/autoproj/workspace.rb
+++ b/lib/autoproj/workspace.rb
@@ -228,6 +228,11 @@ module Autoproj
                     manifest.vcs = VCSDefinition.from_raw(
                         type: 'local', url: config_dir)
                 end
+
+                if config.source_dir && Pathname.new(config.source_dir).absolute?
+                    raise ConfigError, 'source dir path configuration must be relative'
+                end
+
                 os_package_resolver.prefer_indep_over_os_packages = config.prefer_indep_over_os_packages?
                 os_package_resolver.operating_system ||= config.get('operating_system', nil)
             end
--- a/lib/autoproj/workspace.rb
+++ b/lib/autoproj/workspace.rb
@@ -191,6 +191,20 @@ module Autoproj
             config.set 'build', path, true
         end
 
+        # (see Configuration#source_dir)
+        def source_dir
+            if config.source_dir
+                File.expand_path(config.source_dir, root_dir)
+            else
+                root_dir
+            end
+        end
+
+        # Change {#source_dir}
+        def source_dir=(path)
+            config.set 'source', path, true
+        end
+
         def log_dir
             File.join(prefix_dir, 'log')
         end
--- a/lib/autoproj/workspace.rb
+++ b/lib/autoproj/workspace.rb
@@ -306,7 +306,7 @@ module Autoproj
                 io.puts "workspace: \"#{root_dir}\""
             end
 
-            Autobuild.srcdir = root_dir
+            Autobuild.srcdir = source_dir
             Autobuild.logdir = log_dir
             if cache_dir = config.importer_cache_dir
                 Autobuild::Importer.default_cache_dirs = cache_dir
@@ -552,7 +552,7 @@ module Autoproj
                 end
 
             pkg = manifest.find_autobuild_package(pkg_name)
-            pkg.srcdir = File.join(root_dir, srcdir)
+            pkg.srcdir = File.join(source_dir, srcdir)
             if pkg.respond_to?(:builddir)
                 pkg.builddir = compute_builddir(pkg)
             end
--- a/test/test_workspace.rb
+++ b/test/test_workspace.rb
@@ -473,6 +473,23 @@ module Autoproj
                 ws.source_dir = 'src'
             end
         end
+
+        describe "#load_config" do
+            attr_reader :ws
+            before do
+                @ws = ws_create
+                FileUtils.mkdir_p File.join(ws.root_dir, '.autoproj')
+                FileUtils.touch File.join(ws.root_dir, '.autoproj', 'config.yml')
+            end
+
+            it "raises if config.source_dir is not relative" do
+                flexmock(Autoproj::Configuration).new_instances.
+                    should_receive(:source_dir).and_return(ws.root_dir)
+                assert_raises(ConfigError) do
+                    ws.load_config
+                end
+            end
+        end
     end
 end
 
--- a/test/test_workspace.rb
+++ b/test/test_workspace.rb
@@ -448,6 +448,31 @@ module Autoproj
                 end
             end
         end
+
+        describe "#source_dir" do
+            attr_reader :ws
+            before do
+                @ws = ws_create
+            end
+
+            it "returns root_dir if 'source' config option is unset" do
+                flexmock(ws.config).should_receive(:source_dir).
+                    and_return(nil)
+                assert_equal ws.source_dir, ws.root_dir
+            end
+
+            it "returns root_dir/source if 'source' config option is set" do
+                flexmock(ws.config).should_receive(:source_dir).
+                    and_return('src')
+                assert_equal ws.source_dir, File.join(ws.root_dir, 'src')
+            end
+
+            it "sets 'source' config option" do
+                flexmock(ws.config).should_receive(:set).
+                    with('source', 'src', true).once
+                ws.source_dir = 'src'
+            end
+        end
     end
 end
 


doudou:
I'd like a lot more rationale for this ...

I tend to prefer to not over-configure things. This fails straight onto the "too over-configuring".

g-arjones:
1. Looks better and makes things more organized within the workspace (autoproj already allows a common build tree through `build´ config option). Having a 'source' common root  like 'build' and 'install' makes sense, IMO;

2. This is part of my ongoing effort of making autoproj and ROS/catkin colaborate

I would be fine with having this in the layout:

```
- src/:
  - package1
  - package2
```

But that causes autoproj to create separate prefixes and an extra sub folder in the build tree

doudou:
> Looks better and makes things more organized within the workspace

Yeah, so the thing with "personal preferences" is that it is what it is ... personal. I spent years adding optional behaviors to a lot of SW because everyone was coming up with their pet personal preference. Not anymore. So :-1: on this front. The major use-case for autoproj is to actually edit files, that's why source code is front and center.

Now for point 2: what's the link with catkin ?

doudou:
As for why there's the ability to have a separate build dir: it so that you can **separate** build, prefix and source, especially for backup reasons or to be able to delete the whole build tree easily. Not because of cosmetic reasons.

You could bootstrap in `src/` and setup your prefix and build dir to ${root}/../prefix and ${root}/../build for the exact same effect.

g-arjones:
> separate build, prefix and source, especially for backup reasons

This definitely improves source separation and eases backup/sharing

> Now for point 2: what's the link with catkin ?

It is a convention within ROS and a requirement of catkin to have source packages together in a common source folder

> You could bootstrap in src/ and setup your prefix and build dir to ${root}/../prefix and ${root}/../build for the exact same effect.

How come? AFAICT, [autoproj joins prefix_dir and the layout to define each package prefix folder](https://github.com/rock-core/autoproj/blob/master/lib/autoproj/workspace.rb#L528)

```
- src/:
   - package
```
If `prefix` config option is `install`, then the package prefix is ´/home/arjones/autoproj_ws/install/src/package`

g-arjones:
> How come? AFAICT, autoproj joins prefix_dir and the layout to define each package prefix folder

Scratch that, I got it wrong. You meant to have the autoproj workspace as a sub folder of the "real" workspace. That would feel "awkward" to say the least.

doudou:
> It is a convention within ROS and a requirement of catkin to have source packages together in a common source folder

Common source folder, or common `src` folder ? Autoproj already places all the packages in a common source folder.

Last time I used ROS, `packages` were anything that had a `manifest.xml` (and then `package.xml`) under a certain root folder. Not the case anymore ?

> This definitely improves source separation and eases backup/sharing

*Separating does*. How you separate is a matter of personal preference ! Your way is only one way.

> How come? AFAICT, autoproj joins prefix_dir and the layout to define each package prefix folder

Not what I meant. (You can also see that I don't overly advertise the layout system ... it really is a relic of the age where I tried to please everyone, and see how much complexity stems from it).

~~~
mkdir -p dev/flat_fish/src
cd dev/flat_fish/src
ruby autoproj_bootstrap
~~~

and in your `~/.autoprojrc` or in the workspace's 'init.rb'

~~~
Autoproj.workspace.prefix_dir = File.expand_path('../prefix', Autoproj.workspace.root_dir)
Autoproj.workspace.build_dir = File.expand_path('../build', Autoproj.workspace.root_dir)
~~~

doudou:
(Oh, and if you want the cleanest of the clean, set `separate_prefixes` to true. Each package gets its own prefix under `prefix_dir`. I don't advertise it too much, because builds usually fail because of shitting environment handling and/or missing dependencies.

g-arjones:
> Oh, and if you want the cleanest of the clean, set separate_prefixes to true. Each package gets its own prefix under prefix_dir

I found that while I was implementing this but that's the opposite of the goal. Has the same effect of the layout system.

> Not what I meant.

Yes, I got what you meant just after I made the comment. What you are suggesting feels awkward, hacky and makes it more difficult to manage and maintain (spreads configuration).

g-arjones:
> Common source folder, or common src folder ? Autoproj already places all the packages in a common source folder.

`src`

doudou:
> What you are suggesting feels awkward, hacky and makes it more difficult to manage and maintain (spreads configuration).

I really don't see how it gets more or less hacky than having a `source_dir` configuration parameter that, in the end, is also hidden from view (noone edits or sees config.yml). It builds on top of existing attributes that can already be set, instead of having yet-another-configuration-parameter. Moreover, to expose the functionality, you will have to either add yet one-more-parameter to the bootstrap to pass it, or edit `init.rb` to set the parameter.

The bottom line is: it's not meant to be used by everyone, because what is meant to be used by everyone is something that makes all Autoproj workspaces look alike. The primary part of "the workspace" in this case being "where the source code is", because that's what one interacts the most with. What you want is to essentially break that, so forgive me if I'm really reluctant to do so.

> > Common source folder, or common src folder ? Autoproj already places all the packages in a common source folder.

> `src`

Yeah ... I should have guessed. How is this enforced in the first place ?

g-arjones:
Just another apoint: this doesn't introduce any new behavior in autoproj and barely adds any complexity (definitely nothing like the layout system you created to please people). It's just a config option that I doubt any one will even notice (as unobtrusive as the separate prefix thing).

doudou:
> Just another apoint: this doesn't introduce any new behavior in autoproj

Well, yes, it does.

> Just another apoint: this doesn't introduce any new behavior in autoproj and barely adds any complexity 

"Small, unobtrusive, __optional__ changes of behaviors" is death of the maintainer by a thousand cuts. I'm already thinking about things or tooling that would break for the people that would use this function because the tooling assumes `source_dir == root_dir`. You could argue that "if you use this you're on your own", but that's never the case. I, as the autoproj maintainer, would take the blame.

Anyways. I really want to find something that works for both of us, I really do. I do appreciate your contributions to the Rock tooling *a lot*. Let's summarize the various possibilities, maybe something will show up ...

As I see it on the ROS side, the issue is to get all ROS packages under a separate folder that is separated from the build and prefix, but with the constraint that packages should not change their names (or the package.xml won't load properly).

There's the layout, but it leads to having separated prefixes, which you don't like.

One cannot just add a `ros/` to the ROS package names, as the names will then be different (obviously) and package.xml won't resolve properly.

There is the `move_package` feature which was introduced for this purpose (packages declared in `orocos-toolchain` that don't have a Rock category and need to be installed in `tools/`). However, the current behavior - that is, that the package ends up in `${prefix}/${original_package_name}` - could be considered a bug. Given that it's a bug we lived with for years, it's unlikely that I will change that behavior anyways. Ideally, under this option, one would put the packages under something descriptive like `ros/`, so there's still the need to call catkin_build with the right option (is autoproj or the user calling this) ? Moreover, the `move_package` is transparent: one does not need to bootstrap with some obscure option to start pulling ROS packages. `move_package` is called straight from within the package set.

From a workspace organization standpoint, and with the idea in mind that one would mix ROS and Rock packages, `move_packages` interesting aspect is that one can keep the package.xml as-is while having a Rock-compatible layout (i.e. with packages organized). Without this, if you want to use subfolders, you are on your own. However, to fix the problem you have with catkin you would need to separate ROS and non-ROS packages, which is suboptimal. )

Bootstrapping in `src/` and separating build and prefix outside of `src/`. The part that I like about this - and the reason I implemented it this way in the first place - is that it shows that you can move your prefix and build folders anywhere. E.g. on a different type of storage, or on a different subvolume (with btrfs), which reduces the snapshot sizes, ... In fine, the build and prefix folders are things you barely touch, so moving them is a lot less disruptive than moving the source folder. By the way, if you restrict backups/snapshots to `src/`, you're missing the whole autoproj configuration - not the nicest. I can tell, I had a bunch of btrfs filesystems dying on me and had to get things out of backup.

The `source_dir` option, that moves the source tree at an arbitrary location under the root of the workspace.

In summary:
- `layout`: separates prefixes
- renaming: breaks `package.xml`
- move_package: works, but will lead to different source trees between ROS and Rock packages. Is done at the package set level, so there's no need to bootstrap in a special way to start using the ROS packages.
- bootstrap the workspace under `src/` and then setup prefix/build in ../build and ../prefix: you don't like
- `source_dir` option: I don't like the idea of adding yet-another-optional behavior in autoproj

Is that a good summary ?

doudou:
(Apropos the installed package.xml being found by catkin, you can create a CATKIN_IGNORE file somewhere to have catkin ignore the folder)

g-arjones:
>  I'm already thinking about things or tooling that would break for the people that would use this function because the tooling assumes source_dir == root_dir

The only context is see this happening is if the tool trying to map package names to source directories (which is wrong, obviously). These tools are already broken by the move_package API anyway..

> There is the move_package feature

That's what I am doing right now, the problem is that the logic to move all packages has to be spread across the several package sets and potentialy in the build configuration. Potentially, we could have a `ros_package` helper in the package set dsl that would create the cmake_package and move the package but would still endup with ROS and Rock/plain cmake packages in a different tree.

> Without this, if you want to use subfolders, you are on your own

With this patch I can mix Rock and ROS packages. Rock packages will be organized just as one would expect, except that it will be in root_dir/src rather than root_dir.

> Is that a good summary ?

Yes, indeed. I guess we agree the best two options are either having "build" and "prefix" outside of the autoproj workspace or this patch. Unfortunately, we don't seem to agree which one is the best.

> you can create a CATKIN_IGNORE file somewhere

Yeah.. That one also

doudou:
Unfortunately, when doing this summary, the whole `one has to bootstrap with source_dir=src to build ROS packages` is basically saying "if we implement it this way, this will eventually become the default".

A.k.a. I just went from "mild bad feeling to very strong dislike".

One worthy thing would be to check ROS2's new build tool, and especially verify whether it still has the same constraint. That would at least inform whether this is something that will disappear as catkin is phased out, or not.

With CATKIN_IGNORE, the argument that prefix dir must be outside the tree doesn't  hold anymore. Assuming that we create a CATKIN_IGNORE in install/, would it be possible to pass the root dir to catkin_build and have catkin be happy ?

A common way to introduce this kind of features is to create a new package set with specific configuration and helpers. One would *definitely* add a `ros_package` helper there, so for me this is not an issue.

*Ideally*, the best option would be the one where one person can have a normal Rock workflow and start adding ROS packages transparently. In the light of this, honestly, it's either move_package or something totally different.

g-arjones:
> check ROS2's new build tool, and especially verify whether it still has the same constraint.

It does. It seems the goal of the new tool is to make it easier to extend (i.e add support for other package types)

> Assuming that we create a CATKIN_IGNORE in install/, would it be possible to pass the root dir to catkin_build and have catkin be happy ?

I suppose we could. I would be a bit sad with this workaround and would very much regret it if creating CATKIN_IGNORE files became part of my workflow. _But_, I guess I could live with it.

> One would definitely add a ros_package helper there

A `ros_package` in the package set is not a good idea, IMO. When working with ROS, I see one having at least 2-3 packages that might contain ROS packages definitions (a common ROS package set, a rosdistro specific package set and a project specific package set). Copying the helper over across several package sets is not an option.

> it's either move_package or something totally different.

move_package was my first try until I found out that moving all packages is not possible (one can not move the packages from within overrides.rb, after they're all defined). Maybe we can change this?

doudou:
> it if creating CATKIN_IGNORE files became part of my workflow.

Who said it should become part of your workflow ? The package set can create it automatically in its `init.rb`.

> A ros_package in the package set is not a good idea, IMO. When working with ROS, I see one having at least 2-3 package sets that might contain ROS packages definitions (a common ROS package set, a rosdistro specific package set and a project specific package set). Copying the helper over across several package sets is not an option.

And all these package sets will depend on a common package set that defines the helpers and do the rest of the setup.

And then I bash my head on the table because I'm dumb ... This package set already exist `https://github.com/rock-ros/package_set/`

> move_package was my first try until I found out that moving all packages is not possible (one can not move the packages from within overrides.rb, after they're all defined). Maybe we can change this?

We could try (a.k.a. I'm definitely not against it). The problem is that srcdir/builddir are computed before loading overrides.rb by design - so that overrides.rb can change them. Not sure how easy it would be to change. Moreover, not an issue with the `catkin_package` helper.

g-arjones:
> The package set can create it automatically in its init.rb

Yes, I just fear I will sometimes have packages in my workspace I don't want built. But fine,.. Let's move on from this, I guess source_dir is really off the table at this point.

> And all these package sets will depend on a common package

How do I `require` a something that is defined in a different package set? Is there an api for that in autoproj or do I have to expand __FILE__ paths?

> And then I bash my head on the table

Do it one more time for my sake. I had to do a lot of the work already done there because I didn't know someone else was doing that as well. :P

> The problem is that srcdir/builddir are computed before loading overrides.rb by design - so that overrides.rb can change them

I will have a look if you are ok with this change. My understanding is that overrides.rb have to be loaded after all package sets are loaded but not necessariily before setup_package_directories is called. Is that correct? In this case, it might be feasible.

> Moreover, not an issue with the catkin_package helper.

Yes, it is. I don't want a tree for ROS packages and a different one for plain cmake/Rock packages.

doudou:
> How do I require a something that is defined in a different package set? Is there an api for that in autoproj or do I have to expand FILE paths?

You don't. These "utility" package sets usually define/load what they provide within their init.rb, and the others get it this way.

> I will have a look if you are ok with this change. My understanding is that overrides.rb have to be loaded after all package sets are loaded but not necessariily before setup_package_directories is called. Is that correct? In this case, it might be feasible.

Well ... not really correct, no. overrides.rb may look at a package's various directories (this is by design), and must therefore be already set then. [Example in rock.core](https://github.com/rock-core/package_set/blob/4426cfa659ebd71bba5b48f9d070da491218093b/overrides.rb#L91). For this very reason, I don't think you actually can do what you want, because it would make processing of `srcdir` in overrides.rb ordering-sensitive (i.e. probably break some other overrides.rb).

> >  Moreover, not an issue with the catkin_package helper.
> Yes, it is. I don't want a tree for ROS packages and a different one for plain cmake/Rock packages.

OMG ... now I get what you want to do ... You really are a stubborn one ...

Here's the thing: what you are trying to do is basically work around the way Autoproj works to make it fit the way *you* think it *should* work. As I always tell other people: you're free to do so, and being able to `move_package` in `overrides.rb` doesn't sound like a bad thing to me (so I'll let you have that if you manage). However, don't complain when it breaks down the line. Autoproj won't _support_ this.

Especially since the way you believe it "should" work is basically there because you want to force Autoproj to follow Catkin's choices, because you want to build catkin packages. These are two different tools with different choices, having imperfect integrations in this case is part of life. Live with it.

g-arjones:
> You don't. These "utility" package sets usually define/load what they provide within their init.rb, and the others get it this way.

Sounds about right.

> OMG ... now I get what you want to do ... You really are a stubborn one ...

Hmmm.. Thought you knew that by now. :-P

> However, don't complain when it breaks down the line. Autoproj won't support this.

Not very encouraging, is it? To be honest, the idea doesn't sound right (setting everything just to move everything a few lines of code later). Honestly, this patch still makes sense and I really don't get why this is such a big deal. Autoproj has quite a few unused, hidden, experimental options that can only be used in combination, this one would probably be the most harmless and the one with a legit use case.

Also, I don't think this is changing the way autoproj works. Moving packages to a different tree is already possible through the layout system (with a few quirks and bug prone, specially during first full builds due to a constantly changing environment).

I guess I will have to live with a fork :(

doudou:
> I guess I will have to live with a fork :(

Suit yourself, I guess. You know that this is basically "my way, or the highway" ? Not very encouraging either.

My problem is: with this, the *only* way that one will ever build catkin packages is if one bootstraps with `source_dir=src/`. 

Which means that: the `source_dir=src/` will _in fine_ become the standard way of bootstrapping a Rock install, just because people will think "but what if I need it ?" (a fair question). You are *not* introducing a "small, optional configuration value". By creating a dependency on it, you basically are asking to change medium-term the way Autoproj workspaces work. *This* is what made me go from "don't really like it" into "full-blown against".

g-arjones:
> You know that this is basically "my way, or the highway" ?

The whole discussion exists because that's what I was trying to avoid in the first place. I use and contribute to Autoproj because I like it, I do it in my spare time and I really appreciate all the ideas and suggestions but unfortunately we could not come up with anything that didn't feel like a hack/workaround.

I'm willing to do the necessary work to implement it in a way you are fine with but it doesn't make any sense  to do it or rely on it knowing that it is likely to break in the near future because Autoproj won't support it.

doudou:
OK .. so.

Just built moveit, using catkin itself and `--source $root_dir  --build $prefix/ros/build`. This worked just fine as soon as I touched `$prefix/CATKIN_IGNORE`. In any case, building with autoproj as it is, is a lost cause since a lot of the ROS repositories contain more than one package - a workflow that Autoproj supports horribly badly[1], so one will have to go running catkin (and have a special package type).

And OK, you tired me enough. I have enough stress as it is. If this is sooooo important to you that you actually would prefer forking autoproj, I guess I'm the one that has to move.

PR #185, opened 2018-03-27
unify expansions between imports and version_control fields

doudou:
The imports were weirdly not expanding some stuff that was expanded
in the version control fields. Refactor the code to use the same code
path, and expand the same stuff. This requires loading constants before
imports, but there is nothing from the constants that can be changed
by parsing the imports


g-arjones:
Looks good :+1:

PR #188, opened 2018-03-29
Report elapsed time

g-arjones:


lib/autoproj/reporter.rb
```diff
@@ -49,8 +49,25 @@ def error(error)
...
         end
+
+        def reset_timer
+            @timer_start = Time.now
+        end
+
+        def elapsed_time
+            return nil unless @timer_start
+            secs = Time.now - @timer_start
+            [[60, :seconds], [60, :minutes], [24, :hours], [1000, :days]].map do |count, name|
```
doudou: I think you might want to read more about [Time#strftime](https://ruby-doc.org/core-2.2.0/Time.html#method-i-strftime)

doudou: Right ... you can't if yout want to achieve the same result (that is, not showing the time periods that are irrelevant). Sorry for the noise.

g-arjones: No problem... :+1:

lib/autoproj/reporter.rb
```diff
@@ -49,8 +49,28 @@ def error(error)
...
+                    secs, n = secs.divmod(count)
+                    next if (val = n.to_i) == 0
+                    "#{val} #{val > 1 ? name + 's' : name}"
+                end
+            end.compact.reverse.join(' ')
+        end
+
         def success
-            Autoproj.message("Command finished successfully at #{Time.now}", :bold, :green)
+            elapsed_string = elapsed_time ? " (took #{elapsed_time})" : ''
```
doudou: nil.to_s is the emptry string. (and ?: is very much un-rubyish). A better way (as in more rubocopyish) would be to just assign ... if 

~~~
elapsed_string = " (took #{elapsed_time})" if elapsed_time
~~~

doudou: (no need to change, just for future reference)

g-arjones: Hmm.. This is new to me.

```ruby
if false
    foo = 'bar'
end
```

I thought `foo` would be unset, in which case `puts foo` would raise `NoMethodError` but Ruby conveniently sets `foo` to `nil` even though the expression is false. Pretty neat :+1:

doudou: In Ruby, local variables are defined after their first assignation, i.e. they are defined (and initialized to `nil`) in *all lines of the same scope* as soon as the parser sees them assigned.

The rule sounds really weird, but it actually makes a lot of sense in the code.

g-arjones: > makes a lot of sense in the code.

Definitely. Thanks!

PR #190, opened 2018-04-03
fix parsing description with html tags in ros manifests

g-arjones:


doudou:
Shouldn't we rather check if whether the tag is at toplevel (i.e. maintain a tag level by balancing the tag_start and tag_end calls, and parse only the tags that have a level of 1) ?

g-arjones:
Not sue... If I got your point, that would allow nested tags within any tag, right? Do we want that?  What do you think?

I am also wondering if you would like to extend this to Rock manifests as well. Currently, a NoMethodError is raised (which may be considered a bug)

doudou:
> I am also wondering if you would like to extend this to Rock manifests as well. Currently, a NoMethodError is raised (which may be considered a bug)

Definitely. I would simply ignore all tags that are not at toplevel within the main manifest parser (maybe simply have the base class forward `tag_begin`/`tag_end` to separate `toplevel_tag_begin`/`toplevel_tag_end` methods, which would allow the current parser code to remain essentially unchanged.

g-arjones:
Updated.

PR #196, opened 2018-04-25
cli: in 'status' do not attempt to snapshot a package that is up to date

doudou:


lib/autoproj/cli/status.rb
```diff
@@ -85,7 +85,17 @@ def status_of_package(package_description, only_local: false, snapshot: false)
                 elsif !File.directory?(pkg.srcdir)
                     package_status.msg << Autoproj.color("  is not imported yet", :magenta)
                 else
-                    if importer.respond_to?(:snapshot)
+                    begin status = importer.status(pkg, only_local)
+                    rescue Interrupt
+                        raise
+                    rescue Exception => e
```
g-arjones: Isin't rescuing ::StandardError alone better than rescuing Exception + Interrupt and reraising Interrupt?

g-arjones: Rescue Exception rescues even stuff like syntax errors..

doudou: > rescues even stuff like syntax errors..

I saw that ... I got used to it because the most exception-heavy code I wrote, Roby, has to isolate *all* exceptions between the framework and the user code. Even NoMethodError. Then, I ended up never questioning the practice.

Not an excuse, mind you, but an explanation.

doudou:
Updated

PR #197, opened 2018-04-25
fix tests after the change in autobuild from \r to ANSI codes to clear lines

doudou:
Depends on:
- [x] https://github.com/rock-core/autobuild/pull/53


g-arjones:
:+1:

doudou:
> Actually, there are more tests failing due to the change in the output.

These are fixed when the autobuild PR is applied.

PR #198, opened 2018-04-27
load package.xml only when enabled explicitely

doudou:


g-arjones:
Can't test right now but looks good to me. One thing though: could we bring back the warning if the package has both package.xml and manifest.xml (and the `use_package_xml` is enabled, obviously)?

doudou:
> and the use_package_xml is enabled, obviously ?

I wouldn't mind. But right now, if `use_package_xml` is set, the only thing that can be loaded is the package.xml, so I'm not sure about the rationale behind the warning.

g-arjones:
Maybe the user might expect manifest.xml to be loaded instead of package.xml and have no clue why things don't work? Not sure though, I'm fine without the warning as well..

doudou:
I have the impression that it might end generating warnings that one cannot get rid of - if one does want to use the `package.xml` even in presence of `manifest.xml`, then the warning will be there forever.

g-arjones:
Fair enough.. Feel free to merge

doudou:
Released as 2.7.0 along with the Python package support (autobuild 1.14.0)

PR #201, opened 2018-06-26
smoothly switch to a XDG layout for global files

doudou:
The code should pick a XDG location for new files/folders, but keep
the current locations when they exist.

### DIFF
--- a/lib/autoproj/ops/install.rb
+++ b/lib/autoproj/ops/install.rb
@@ -146,10 +146,19 @@ module Autoproj
             def gems_install_path=(path)
                 @gems_install_path = path
             end
+
+            private def xdg_var(varname, default)
+                if (env = ENV[varname]) && !env.empty?
+                    env
+                else
+                    default
+                end
+            end
+
             # Install autoproj in Gem's default user dir
             def install_gems_in_gem_user_dir
-                xdg_default_gem_path = File.join(
-                    Dir.home, '.local', 'share', 'autoproj', 'gems')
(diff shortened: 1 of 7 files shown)


lib/autoproj/ops/install.rb
```diff
@@ -52,8 +52,16 @@ def initialize(root_dir)
                 @ruby_executable = config['ruby_executable']
                 @local = false
 
+                xdg_default_gem_path = File.join(Dir.home, '.local', 'share', 'autoproj', 'gems')
```
g-arjones: Why isn't this XDG_DATA_HOME instead? Same for the bootstrap/install scripts..

Ref: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

doudou: This is XDG_DATA_HOME.

Unfortunately, the xdg gem actually does not handle the environment variables - it only handles the default XDG folders in Dir.home. Moreover, install.rb should be dependency-free (it ends up included in the install/bootstrap script) and therefore cannot use the xdg gem.

doudou: > Unfortunately, the xdg gem actually does not handle the environment variables

Well .. this is actually not true.

doudou: ... and I therefore added support for the XDG_DATA_HOME env var.

g-arjones:
Nice.. Looks good to me :+1:

PR #202, opened 2018-06-26
Global workspace registration

doudou:
Depends on:
- [ ] #201 

This commit is inspired by vagrant's global-status command, which allows to have an overview
of the vagrant environments currently active in any given system. This PR introduces the `global`
subcommand, and in particular `global status`, which tells which workspaces are available and where
they are.

This is really for bookkeeping. It for instance allows to find out that some build/prefix directories
from old workspaces are still there.

### DIFF
--- a/lib/autoproj/workspace.rb
+++ b/lib/autoproj/workspace.rb
@@ -1,3 +1,6 @@
+require 'autoproj/ops/loader'
+require 'xdg'
+
 module Autoproj
     class Workspace < Ops::Loader
         attr_reader :root_dir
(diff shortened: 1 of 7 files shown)


lib/autoproj/workspace.rb
```diff
@@ -457,24 +461,89 @@ def load_main_initrb
...
+                end.compact
+            else
+                []
+            end
+        end
+
+        def self.save_registered_workspaces(workspaces)
+            workspaces = workspaces.map do |w|
+                Hash['root_dir' => w.root_dir,
+                     'prefix_dir' => w.prefix_dir,
```
g-arjones: Is it a good idea to have this workspace-specific configuration stored globally? I kinda dislike the fact that this will probably be left outdated if the workspace's config.yml changes specially if the user has to manually register and re-register (to update) this information. I would instead keep only the `root_dir` in the global registry and load `prefix_dir` and `build_dir` from workspace's config.yml (for `autoproj global status`).

I also wonder if workspaces shouldn't be registered automatically (maybe whenever env.sh is updated?)

doudou: It actually [is registered automatically](https://github.com/rock-core/autoproj/pull/202/files#diff-eae79889f58431d9f820a91c7210cfd8R292). The register command was there more for testing purposes. It could be removed if you believe it's better.

The point of saving the various folders it to be able to track stale folders once workspaces get deleted (for people like me that separate build, install and source).

g-arjones: > It could be removed if you believe it's better

No, automatically is cool.

> The point of saving the various folders it to be able to track stale folders once workspaces get deleted (for people like me that separate build, install and source).

Makes sence.. I missed that use case.

g-arjones:
Looks good ! :+1:

PR #212, opened 2018-08-08
Post phase hooks

doudou:
These hooks allow plugins to execute code after overall operations
such as build and update.

This is introduced for the [autoproj-sync plugin](https://github.com/rock-core/autoproj-sync), so that it can [automatically synchronise after a build or update](https://github.com/rock-core/autoproj-sync#automated-synchronisation)

See the documentation of register_post_command_hook for the list
of available hooks and their documentation

### DIFF
--- a/lib/autoproj/cli/main.rb
+++ b/lib/autoproj/cli/main.rb
@@ -3,6 +3,7 @@ require 'tty/color'
 require 'autoproj/cli/main_test'
 require 'autoproj/cli/main_plugin'
 require 'autoproj/cli/main_global'
+require 'autoproj/cli'
 require 'autoproj/reporter'
 
 module Autoproj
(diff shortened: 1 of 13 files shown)


lib/autoproj/cli/main.rb
```diff
@@ -3,13 +3,20 @@
...
+require 'autoproj/cli'
 require 'autoproj/reporter'
 
 module Autoproj
     module CLI
         def self.basic_setup
             Encoding.default_internal = Encoding::UTF_8
             Encoding.default_external = Encoding::UTF_8
+
+            argv = ARGV.find_all { |arg| arg != "--no-plugins" }
```
g-arjones: why can't we have this integrated in the Thor interface?

doudou: Because plugins extend the thor interface (they add commands & subcommands). So, need to be loaded before you start executing Thor

g-arjones:
:+1:

PR #216, opened 2018-09-04
fix some more corner-cases related to moving bundler in .local/

doudou:
The main issue is that Gem would pick up a bundler already installed
in $HOME/.gem, which would work until we get to our shims, which
reset GEM_HOME and GEM_PATH to ensure an isolation of the Autoproj
workspace gems.

Make sure the bundler we find is indeed in #gems_gem_home, and
that it gets installed there regardless of the state of the rest
of the system.

### DIFF
--- a/bin/autoproj_bootstrap
+++ b/bin/autoproj_bootstrap
@@ -50,8 +50,6 @@ module Autoproj
                 @env = Hash.new
                 env['RUBYOPT'] = []
                 env['RUBYLIB'] = []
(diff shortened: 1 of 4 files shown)


g-arjones:
Yep.. Was just going to comment about what these last commits fix. Looks good to me now :+1:

doudou:
Thanks

AlexanderFabisch:
I think this is the cause for the bug in #217

PR #221, opened 2018-10-26
Run tests recursively

### LINKED ISSUE #220: autoproj test exec does not run all enabled tests
@doudou would you please try to reproduce this?

g-arjones:
Fixes #220


doudou:
My CI actually runs tests separately, so never caught this. Thanks !

PR #225, opened 2018-11-19
implement apt repository manager

g-arjones:
This is a work in progress. 

@doudou I would like to get your thoughts before proceeding with this.. I am still deciding how to represent repositories in an `*.osrepos` file within a package set. A couple of ideas:

1. 
```yaml
- ubuntu:
  - xenial:
    type: repo
    repo: 'deb http://br.archive.ubuntu.com/ubuntu/ xenial universe'

- ubuntu:
  - xenial:
    type: key
    keyserver: 'hkp://ha.pool.sks-keyservers.net:80'
    id: '36A1D7869245C8950F966E92D8576A8BA88D21E9'
```

2.
```yaml
- type: repo
  repo: 'deb http://br.archive.ubuntu.com/ubuntu/ xenial universe'
  applies_to:
    - ubuntu:
      - xenial

- type: key
  keyserver: 'hkp://ha.pool.sks-keyservers.net:80'
  id: '36A1D7869245C8950F966E92D8576A8BA88D21E9'
  applies_to:
    - ubuntu:
      - xenial
```

Thoughts?


planthaber:
Hi, i like the idea, perhaps this could also be used for the rock debian packages @2maz?
As the ubuntu version is part of the repo url anyways, i'd propose this:

As the debian wiki states that the sources entry consists of:
* Archive type
* Repository URL
* Distribution
* Component

Source: https://wiki.debian.org/SourcesList

I'd say you should stick to that. Perhaps like this:

```
- ubuntu:
    type: apt
    url: 'http://br.archive.ubuntu.com/ubuntu/'
    distribution: xenial
    components: universe
```

The same scheme could also be used for other apt-based distributions. e.g. debian:
```
- debian:
    type: apt
    url: 'http://deb.debian.org/debian'
    distribution: stretch
    components: free
```

and later be extended to other package managers

g-arjones:
@planthaber thanks for contributing the discussion..

What you are suggesting was my first idea but then I didn't think it was ideal for a few reasons:

1. Despite the fact that the field is called "distribution" it is often used for other purposes (it is common to have things like "stable" and "security" in this field) so we still have to specify the distribution that the entry applies to;
2. There are cases that the same entry applies to multiple distributions and/or multiple releases;
3. What I suggested allows things like `ppa:libreoffice/libreoffice-5-0`

What do you think?

planthaber:
Hi, hre my answers:

1. True, but "stable" and "security" are alst treaded as distribution by apt, so i think it is ok to do so.
2. No, the first param after the url ist the distribution, the rest are components. Here ist the part from here: https://wiki.debian.org/SourcesList

>deb http://site.example.com/debian distribution component1 component2 component3
>deb-src http://site.example.com/debian distribution component1 component2 component3

>Archive type
>
>The first word on each line, deb or deb-src, indicates the type of archive. Deb indicates that the archive contains binary packages (deb), the pre-compiled packages that we normally use. Deb-src indicates source packages, which are the original program sources plus the Debian control file (.dsc) and the diff.gz containing the changes needed for packaging the program.
>
>Repository URL
>
>The next entry on the line is a URL to the repository that you want to download the packages from. The main list of Debian repository mirrors is located here.
>
>Distribution
>
>The 'distribution' can be either the release code name / alias (jessie, stretch, buster, sid) or the release class (oldstable, stable, testing, unstable) respectively. If you mean to be tracking a release class then use the class name, if you want to track a Debian point release, use the code name.
>
>For example, if you have a system running Debian 9.6 "stretch" and don't want to upgrade when Debian buster releases, use 'stretch' instead of 'stable' for the distribution. If you always want to help test the testing release, use 'testing'. If you are tracking buster and want to stay with it from testing to end of life, use 'buster'.
>
>Component
>
>main consists of DFSG-compliant packages, which do not rely on software outside this area to operate. These are the only packages considered part of the Debian distribution.
>
>contrib packages contain DFSG-compliant software, but have dependencies not in main (possibly packaged for Debian in non-free).
>
>non-free contains software that does not comply with the DFSG. 
>

3. In the end, ppa urls are adding a normal entry in the /etc/apt/sources.list.d/ folder
`deb http://ppa.launchpad.net/libreoffice/libreoffice-5-0/ubuntu bionic main`
another way would be to treat ppas as extra type
```
- libreoffice:
    type: ppa
    url: ppa:libreoffice/libreoffice-5-0
```

g-arjones:
1. Yes, the problem I have with this is that one will often have to write the same thing a couple of times and that can be confusing for the user;
2. I am not sure what you mean by "No". Here are a few examples of entries that are valid for multiple distributions and releases: 

See https://yarnpkg.com/lang/en/docs/install/#debian-stable, 
See https://docs.bazel.build/versions/master/install-ubuntu.html,
See https://www.sublimetext.com/docs/3/linux_repositories.html,
```deb [arch=amd64] http://packages.microsoft.com/repos/vscode stable main```
```deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main```

These are just a few I have in my current setup but I am sure we can find plenty.

3. Both options would be ok but with trade offs: by using the expanded version of the ppa entry you have to also manually add the auth key to the repo and a different `type` field is more stuff to parse/validate.

Do you have a suggestion to deal with 2. specifically?

**Comment by doudou** [comment]
I would stick to having the repo string as the original (1). It's what will end up in the source file, therefore it has all the expressive power needed. Plus, this is what people get from installation instructions and/or source files. Nobody tells people to "get the repo at URL x, distribution y and component z". A.k.a. I very much prefer the (1) of @g-arjones original proposal.

The interest I see with having a separate ppa type is to get the key automatically. Note that internally `type: ppa` could be expanded into key/repo before being processed for installation.

The one thing you should add is the ability to specify the name of the created source file. I've had to use it with Ansible, since some packages install their own source files and this ends up generating warnings because of duplicate entries.

**Comment by g-arjones** [comment]
Ok, cool.. I will carry on with this then.

@doudou Also, could you please check Gitter? :P

**Comment by g-arjones** [comment]
@doudou Took a little while before I could finish this.. Looking forward for your comments

**Comment by annaborn** [comment]
@doudou: could you explain me please, what is the purpose of *.osrepos? In which cases it is used?

**Comment by annaborn** [comment]
@doudou Thank you for link! 
However I undestood, all dependencies rely on the detected os. 
Is there a possibility to use deb package instead of the default os package?
One example would be odb package. For 18.04 we have to use workaround to install proper working version of odb. 
```
 run("prepare", "wget http://ftp.de.debian.org/debian/pool/main/o/odb/odb_2.4.0-10_amd64.deb")
 run("prepare", "sudo dpkg -i odb_2.4.0-10_amd64.deb")
 run("prepare", "rm odb_2.4.0-10_amd64.deb")
```
My question is, would it make sence to implement this mechanism inside autoproj? E.g. to allow the user to add link to dep package inside .osdeps:

```
odb:
    ubuntu:
       '16.04':
           - odb
       '18.04':
            amd64: http://ftp.de.debian.org/debian/pool/main/o/odb/odb_2.4.0-10_amd64.deb
```

**Comment by doudou** [comment]
Oh boy.

I guess it would. What do you think @g-arjones ?

From an implementation standpoint, you would want to have a pattern in place of `odb`
so that the handling is passed on to the AptDpkgManager class, e.g.

~~~
odb:
    ubuntu:
       '16.04':
           - odb
       '18.04':
            amd64: http://ftp.de.debian.org/debian/pool/main/o/odb/odb_2.4.0-10_amd64.deb
~~~

**Comment by g-arjones** [comment]
In principle, it does make sense but I think it's tricky to get right. dpkg does not install dependencies and may leave apt broken if something was missing. We could try to do an `apt install -f` and/or `apt update --fix-missing` whenever a dpkg is installed with this mechanism but I am not sure that's safe/reliable enough to do automatically.

Definitely something that is worth having a look into, IMO...

**Comment by doudou** [comment]
I would definitely avoid `-f` / `--fix-missing` in a non-interactive way, it sometimes decide to remove a LOT of packages you'd rather keep :P.

I'd leave specifying the deps to the user: he/she would have to add the other packages in the same entry (the same way one has to specify APT dependencies within a ruby gem that has some).

**Comment by g-arjones** [comment]
> he/she would have to add the other packages

This is where it gets tricky.., I am not sure if installing a missing package after apt is already broken is enough to fix it (--fix-missing may still be necessary, not sure). Which means you have to guarantee that you install dependencies in the right order

**Comment by doudou** [comment]
> Which means you have to guarantee that you install dependencies in the right order

For me, it only means that you have to the APT packages before the manual dpkg ones. I.e. we would not handle having other APT packages depending on the manually installed one (which in any case would require adding a preference file to avoid having APT upgrade it, yet another level of tricky)

**Comment by g-arjones** [comment]
And what happens if a package that you expect to install manually with dpkg gets installed automatically by apt (i.e due to another apt package depending on it)?

**Comment by planthaber** [comment]
I recently learned, that apt(-get) learned to install .deb packages including its dependencies.
```bash
    sudo apt-get install ./package.deb
```
But is don't now for how long it is around.

Just two ideas Anna and I had discussing this:

* allow .deb file definitions in .osdeps files: like
```yaml
    odb: 
      ubuntu:
         [16.04, 20.04]: odb
         [18.04]: deb: http://....
```
* implementing it as package command for autobuild similar to "archive"
 * source.yml:
```yaml
    odb_deb:
        type: deb
        url: http://...
```
autobuild:
```ruby
deb_package odb_deb
```

  * This way one could install a deb_package and use override_osdep "odb" odb_deb in a buildconf.
  * but it installs to the system, so this would be not so clean

doudou:
Gens, please move this to a new issue.

PR #228, opened 2018-11-25
update debian os packages if a newer version is available

### LINKED ISSUE #227: Autoproj should upgrade OS packages when possible
Specifically, the apt-dpkg package manager should be doing this..

g-arjones:
This is backward compatible....

Closes #227

### DIFF
--- a/lib/autoproj/package_managers/apt_dpkg_manager.rb
+++ b/lib/autoproj/package_managers/apt_dpkg_manager.rb
@@ -17,6 +17,7 @@ module Autoproj
             end
 
             def configure_manager
+                super
                 ws.config.declare 'apt_dpkg_update', 'boolean',
                     default: 'yes',
                     doc: ['Would you like autoproj to keep apt packages up-to-date?']
--- a/lib/autoproj/package_managers/apt_dpkg_manager.rb
+++ b/lib/autoproj/package_managers/apt_dpkg_manager.rb
@@ -11,25 +11,24 @@ module Autoproj
                 @status_file = status_file
                 @installed_packages = nil
                 @installed_versions = nil
-                @ws = ws
                 super(ws, true,
                       %w{apt-get install},
                       %w{DEBIAN_FRONTEND=noninteractive apt-get install -y})
             end
 
             def configure_manager
-                @ws.config.declare 'apt_dpkg_update', 'boolean',
+                ws.config.declare 'apt_dpkg_update', 'boolean',
                     default: 'yes',
                     doc: ['Would you like autoproj to keep apt packages up-to-date?']
                 keep_uptodate?
             end
 
             def keep_uptodate?
-                @ws.config.get('apt_dpkg_update')
+                ws.config.get('apt_dpkg_update')
             end
 
             def keep_uptodate=(flag)
-                @ws.config.set('apt_dpkg_update', flag, true)
+                ws.config.set('apt_dpkg_update', flag, true)
             end
 
             def self.parse_package_status(installed_packages, installed_versions, paragraph)
--- a/lib/autoproj/package_managers/apt_dpkg_manager.rb
+++ b/lib/autoproj/package_managers/apt_dpkg_manager.rb
@@ -8,15 +8,20 @@ module Autoproj
             def initialize(ws, status_file = "/var/lib/dpkg/status")
                 @status_file = status_file
                 @installed_packages = nil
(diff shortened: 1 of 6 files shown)


lib/autoproj/package_managers/apt_dpkg_manager.rb
```diff
@@ -8,23 +8,28 @@ class AptDpkgManager < ShellScriptManager
...
                 @status_file = status_file
                 @installed_packages = nil
+                @installed_versions = {}
                 super(ws, true,
                       %w{apt-get install},
                       %w{DEBIAN_FRONTEND=noninteractive apt-get install -y})
             end
 
-            def self.parse_package_status(installed_packages, paragraph)
+            def self.parse_package_status(installed_packages, paragraph, installed_versions: {})
```
doudou: Try to stick to the current interface, that is `(installed_packages, installed_versions, paragraph)`

lib/autoproj/package_managers/apt_dpkg_manager.rb
```diff
@@ -8,23 +8,28 @@ class AptDpkgManager < ShellScriptManager
...
+                        end
                     end
                     if paragraph =~ /^Provides: (.*)$/
                         installed_packages.merge($1.split(',').map(&:strip))
                     end
                 end
             end
 
-            def self.parse_dpkg_status(status_file)
+            def self.parse_dpkg_status(status_file, installed_versions: {})
```
doudou: Don't add a out argument like this when the method returns its parsed data. Just return both values. I'm guessing that you do this for backward compatibility, but these are meant to be internal APIs.

doudou:
I'm not against the functionality _per se_, but I'm not too sure about it either.

If I want to auto-update, I set up my Ubuntu system to do so. Otherwise, I update the whole system when I want to. I usually don't want tools to do it behind my back.

lib/autoproj/package_managers/apt_dpkg_manager.rb
g-arjones: This was for backward compatibility

lib/autoproj/package_managers/apt_dpkg_manager.rb
g-arjones: Yep, that was the reason. Will refactor if you don't care about that

g-arjones:
Even though I don't consider this being "behind one's back" (the user explicitly calls `update` and autoproj does update some osdeps, i.e: gems), I do get your point. But, I think that it is expected that autoproj update all dependencies, not just "some of them". In fact, I got a few complaints from users specifically mentioning the `--force-uptodate` option that made them think that autoproj kept osdeps up to date. So, as a middle ground I would suggest:

1. We could add a configuration option for this (which should be enabled by default IMO but that's fine if you prefer to keep it off);
2. Warn the user that there are osdeps that should/could be updated and tell he/she how to do it using autoproj if auto update is off.

Thoughts?

doudou:
> Thoughts?

I really don't know ... It feels wrong, but I can't pinpoint what exactly.

planthaber:
I always saw autoproj as a workspace tool to maintain source packages and their dependencies.

I personally linked "update" to updating the workspace, not the OS. In my opinion it should not be the default. While it could be a configuration option, I think it should be a command line flag or autoproj command like.

The osdeps in autoproj are handled seperately from source packages, I think auto-updating osdeps with a update meant for source packages just makes the usage fuzzy. I'd prefer having a 
`$> autoproj osdeps update` command.

Fun fact, that one is already documented: 
> $> autoproj help osdeps
> [--update], [--no-update]            # whether already installed packages should be updated or not

But I think it would be fine if just `autoproj osdeps` in the console triggers the update

g-arjones:
Fair enough... I will implement this in a package set. Thanks for the discussion.

doudou:
>Fair enough... I will implement this in a package set. Thanks for the discussion.

Well ... why ? @planthaber was basically agreeing with you. I think it's a good addition, I'm not sure about what's the best option, that's all.

I would go for (1). You could even ask on bootstrap (the way the osdeps system asks for which systems should be handled by autoproj).

g-arjones:
I got it all wrong then... Sorry about that. I will implement (1) and reopen this.

g-arjones:
Done...

doudou:
Looks good to me

PR #235, opened 2019-01-29
Build status report

sbolivarm:
Added new template to create a json file with a summary of the build process of all packages.


lib/autoproj/ops/build.rb
```diff
@@ -84,7 +86,26 @@ def force_build_packages(selected_packages, all_enabled_packages)
...
+                    Autobuild.apply(all_enabled_packages, "autoproj-build", ['build'], options)
+                ensure
+                    build_report(all_enabled_packages)
+                end
+            end
+
+            TEMPLATES_DIR = File.join(File.dirname(__dir__), 'templates')
+
+            def build_report(package_list)
+                Autobuild.message "INFO: creating build report...."
```
doudou: Autobuild/Autoproj don't prefix their messages with INFO and stuff. Moreover, this would be a debug message - we usually don't emit log messages for this kind of internal processes (otherwise we would get twice as much output)

lib/autoproj/ops/build.rb
```diff
@@ -84,7 +86,26 @@ def force_build_packages(selected_packages, all_enabled_packages)
...
+
+                log_dir = Autobuild.logdir
```
doudou: From within autoproj, the log dir you should consider is `Workspace#log_dir`. The current workspace is accessible in the CLI classes as `#ws`, e.g. `ws.log_dir`

lib/autoproj/templates/build_report_helper.erb
```diff
@@ -0,0 +1,21 @@
+<% @first_line = true -%>
+{
+   "build_report" : {
+       "packages": [<% package_list.each do |pkg_name| -%>
+<% if @first_line == false  %>           },    <% end -%>
+<% @first_line = false  %>
+           {
+               "name": "<%= Autobuild::Package[pkg_name].name.to_s %>",
```
doudou: Within autoproj, use `ws.find_autobuild_package(pkg_name)`. This would be more future-proof. This global registration of packages on `Autobuild::Package` is meant to be phased out eventually.

lib/autoproj/ops/build.rb
g-arjones: Do you really need to print this?

lib/autoproj/templates/build_report_helper.erb
```diff
@@ -0,0 +1,21 @@
...
+               "time": "<%= Time.now.to_s %>",
+               "import_invoked": "<%= Autobuild::Package[pkg_name].import_invoked?.to_s %>",
```
doudou: You don't need the `#to_s` in a `<%= ... %>` template, it's implicit (otherwise all of them would need it anyways ...).

lib/autoproj/ops/build.rb
```diff
@@ -84,7 +86,26 @@ def force_build_packages(selected_packages, all_enabled_packages)
...
+            
+                Autobuild.register_logfile("build_report.json")
```
doudou: Make this name a class-level constant, and provide an accessor for the full path to the file, e.g.

~~~
def build_report_path
   File.join(ws.log_dir, BUILD_REPORT_BASENAME)
end
~~~

lib/autoproj/templates/build_report_helper.erb
g-arjones: To be honest, I don't think a template is the right tool for this... Why not just generate a hash with the data and output it using `JSON#pretty_generate` ? it would be much easier to maintain and  we wouldn't have to worry about the json format in the template file...

lib/autoproj/ops/build.rb
```diff
@@ -84,7 +89,28 @@ def force_build_packages(selected_packages, all_enabled_packages)
...
+            # @return [String,nil] the path, or nil if the report should not
+            #    be generated
+            def report_path
+                File.join(@report_dir, REPORT_BASENAME) if @report_dir
+            end
+
+            def build_report(package_list)
+                erb = File.read(File.join(TEMPLATES_DIR, "build_report_helper.erb"))
+                FileUtils.mkdir_p @report_dir
+                IO.write(report_path, ::ERB.new(erb, nil, '-').result(binding))
```
g-arjones: I think this should be:

```ruby

packages = []
package_list.each do |pkg_name|
  pkg = ws.find_autobuild_package(pkg_name)
  packages << {  name: pkg.name,
                 import_invoked: pkg.import_invoked?,
                 prepare_invoked: pkg.prepare_invoked?,
                 build_invoked: pkg.build_invoked?,
                 failed: pkg.failed?,
                 imported: pkg.imported?,
                 prepared: pkg.prepared?,
                 built: pkg.built?
  }
end

build_report = JSON.pretty_generate({ build_report: { timestamp: Time.now }, packages: [packages] })
```

And no templates...

doudou: Agreed. Should just be `json = package_list.map do ... end` ;-)

g-arjones: Even better :P

g-arjones:
Looks better :+1:

lib/autoproj/ops/build.rb
```diff
@@ -84,7 +89,42 @@ def force_build_packages(selected_packages, all_enabled_packages)
...
+            #    be generated
+            def report_path
+                File.join(@report_dir, REPORT_BASENAME) if @report_dir
+            end
+
+            def build_report(package_list)
+                FileUtils.mkdir_p @report_dir
+
+                packages = package_list.map do |pkg_name|
+                    pkg = Autobuild::Package[pkg_name]
```
doudou: I think you'll have to at least pass a Manifest object to the class in order to avoid accessing Autobuild::Package

doudou: There is no workspace, but there is a manifest already there. Use `manifest.find_autobuild_package` to resolve the packages.

sbolivarm: Agreed. Updating.

lib/autoproj/ops/build.rb
```diff
@@ -84,7 +89,42 @@ def force_build_packages(selected_packages, all_enabled_packages)
...
+            #    be generated
+            def report_path
+                File.join(@report_dir, REPORT_BASENAME) if @report_dir
+            end
+
+            def build_report(package_list)
+                FileUtils.mkdir_p @report_dir
+
+                packages = package_list.map do |pkg_name|
+                    pkg = manifest.find_autobuild_package(pkg_name)
```
doudou: No need for the assignment here. Just `manifest.find_autobuild...`

lib/autoproj/ops/build.rb
```diff
@@ -84,7 +89,42 @@ def force_build_packages(selected_packages, all_enabled_packages)
...
+                        prepare_invoked: pkg.prepare_invoked?,
+                        build_invoked: pkg.build_invoked?,
+                        failed: pkg.failed?,
+                        imported: pkg.imported?,
+                        prepared: pkg.prepared?,
+                        built: pkg.built?
+                    }
+                end
+
+                build_report = JSON.pretty_generate({ build_report: { timestamp: Time.now, packages: packages }})
```
doudou: Please cut lines at 90 characters, e.g.

~~~
build_report = JSON.pretty_generate({
   build_report: {
      timestamp: Time.now,
      packages: packages
   }
})

test/ops/test_build.rb
```diff
@@ -0,0 +1,53 @@
...
+                assert_equal Hash['build_report' => {'timestamp' => Time.mktime(1970,1,1).to_s, 'packages' => []}], json
+            end
+
+            it "works with just one package" do
+                @build.build_report(['pkg1'])
+                json = read_report
+                assert_equal Hash['build_report' => {'timestamp' => Time.mktime(1970,1,1).to_s,'packages' => [expected_package('pkg1')]}], json
+            end
+
+            it "exports the status of the given packages" do
```
doudou: You never test one of the "true" cases....

lib/autoproj/ops/build.rb
sbolivarm: I use the variable pkg to fill the hash object, do you mean to use `manifest.find_autobuild...` in all lines for all properties? 
e.g. replace `name: pkg.name` with `name: manifest.find_autobuild_package(pkg_name).name` ?

doudou: No, you're right. I misread the code. My bad.

test/ops/test_build.rb
sbolivarm: I've added failed and successful packages to the tests.

PR #236, opened 2019-01-30
cli: fix Gemfile generated by `autoproj update <PACKAGE>` when a gem is selected by the manifest

### LINKED ISSUE #233: some Gems are not included in Gemfile when aup is done inside package
When trying to update a package (base/types in my case), I am getting the following error: 

`can't find executable iruby for gem iruby. iruby is not currently included in the bundle, perhaps you meant to add it to your Gemfile? (Gem::Exception)`

When checking the Gemfile, iruby is not part of it. 

If one tries to update from root this error doesn't happen.

doudou:
When calling update with a specific package selection, the codepath
is different than when all packages are selected. In this codepath,
only the osdeps from the selected source packages were passed to
the package managers, which was causing the Gemfile generated for
bundler to be partial.

Closes #233

### DIFF
--- a/lib/autoproj/cli/update.rb
+++ b/lib/autoproj/cli/update.rb
@@ -109,7 +109,8 @@ module Autoproj
                 source_packages, osdep_packages, import_failure =
                     update_packages(
                         selected_packages,
-                        osdeps: options[:osdeps], osdeps_options: osdeps_options,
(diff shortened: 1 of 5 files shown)


g-arjones:
Great fix :+1:

PR #237, opened 2019-02-05
simplify osrepos format

g-arjones:
This changes the format from:

```yaml
- ubuntu:
  - xenial:
    type: repo
    repo: 'deb http://packages.ros.org/ros/ubuntu xenial main'
  - xenial:
    type: repo
    repo: 'deb http://packages.osrfoundation.org/gazebo/ubuntu-stable xenial main'
  - xenial:
    type: key
    keyserver: 'hkp://ha.pool.sks-keyservers.net:80'
    id: 421C365BD9FF1F717815A3895523BAEEB01FA116
  - xenial:
    type: key
    url: 'http://packages.osrfoundation.org/gazebo.key'
    id: D2486D2DD83DB69272AFE98867170598AF249743
```

to:

```yaml
- ubuntu:
  - xenial:
    - type: repo
      repo: 'deb http://packages.ros.org/ros/ubuntu xenial main'
    - type: repo
      repo: 'deb http://packages.osrfoundation.org/gazebo/ubuntu-stable xenial main'
    - type: key
      keyserver: 'hkp://ha.pool.sks-keyservers.net:80'
      id: 421C365BD9FF1F717815A3895523BAEEB01FA116
    - type: key
      url: 'http://packages.osrfoundation.org/gazebo.key'
      id: D2486D2DD83DB69272AFE98867170598AF249743
```

which is less verbose and looks nicer in my opinion..

Given that the whole feature is not released yet, I am shamelessly breaking backwards compatibility...

### DIFF
--- a/lib/autoproj/os_repository_resolver.rb
+++ b/lib/autoproj/os_repository_resolver.rb
@@ -56,16 +56,9 @@ module Autoproj
         def all_entries
             definitions.map do |distribution|
                 distribution.values.map do |release|
-                    release.map do |entry|
-                        remove_identifier_from_entry(entry.dup)
-                    end
+                    release.map(&:values)
                 end
-            end.flatten
-        end
-
-        def remove_identifier_from_entry(entry)
-            entry.delete(entry.keys.first)
-            entry
+            end.flatten.uniq
         end
 
         def entry_matches?(entry, identifiers)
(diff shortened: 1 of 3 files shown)


doudou:
I don't have a problem with the list of repos being an array. This is a different use-case than the osdeps (which have names), so I don't see a problem with having a different format.

The one thing that could be brought closer to the osdeps format is to use hashes for the first two levels ("ubuntu" and "xenial"), the way it's done in osdeps. What do you think ?

g-arjones:
The problem I see with that is a user doing something like:

```
ubuntu:
  xenial:
    - type: repo
      repo: 'deb http://packages.ros.org/ros/ubuntu xenial main'

ubuntu:
  xenial:
    - type: repo
      repo: 'deb http://packages.osrfoundation.org/gazebo/ubuntu-stable xenial main'
```

He/she is probably expecting to be defining two different things but in fact the second one is overriding the first one...

doudou:
> He/she is probably expecting to be defining two different things but in fact the second one is overriding the first one...

I don't have a hard feeling either way. I see the point of having the two formats as close as possible, but I'll go with the flow on this one.

g-arjones:
Also makes merging several .osrepos files trickier...

g-arjones:
> I see the point of having the two formats as close as possible

I do too, which is why I am proposing this change... From where I stand, this is as close as possible (without making it confusing)

doudou:
> (without making it confusing)

My point was: this is a limitation of YAML, and osdeps already are using this (arguably possibly confusing) format, so what the heck

But anyways, good to go on my side.

g-arjones:
Ok then.. I have also opened a PR on the rock-and-syskit website repo

PR #239, opened 2019-02-16
properly finalize package setup when updating with an empty selection

g-arjones:
Some code paths skip package setup during update/import operations, i.e:

```console
$ autoproj update --config
```

and

```console
$ autoproj switch-config branch=master
```

This may have unintended consequences specially in post update hooks. Underlying issue: https://github.com/rock-core/autoproj/blob/6ae7ff7a65c98a2bf9bf2536a0d1d6f0ad01a982/lib/autoproj/cli/update.rb#L96

### DIFF
--- a/lib/autoproj/cli/update.rb
+++ b/lib/autoproj/cli/update.rb
@@ -97,6 +97,8 @@ module Autoproj
                     command_line_selection, selected_packages =
                         finish_loading_configuration(selected_packages)
                 else
+                    ws.setup_all_package_directories
+                    ws.finalize_package_setup
                     command_line_selection, selected_packages = [], PackageSelection.new
                 end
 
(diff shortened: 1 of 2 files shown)


doudou:
It kind of feels ... asymmetric and complicated. What about simply calling `finalize_package_setup` in https://github.com/rock-core/autoproj/blob/6ae7ff7a65c98a2bf9bf2536a0d1d6f0ad01a982/lib/autoproj/cli/update.rb#L100 ?

g-arjones:
Honestly, I am not sure. I tried to understand why it was not already there and couldn't. Given how explicit the check is, my guess was that there was a reason for not doing it there.

Seemed safer to do it at a later stage but these are just assumptions. If you feel it's ok to do what you suggested I totally agree it's way simpler..

doudou:
> If you feel it's ok to do what you suggested I totally agree it's way simpler..

If it fixes the issue you found, that would be best I think.

g-arjones:
@doudou Updated.. 

I think the answer to "why this wasn't done before" is simply "because it was not necessary" before the introduction of post update hooks.

PR #241, opened 2019-02-21
watch: workaround rb-inotify breaking when stopping within a callback

doudou:
See https://github.com/guard/rb-inotify/pull/93

### DIFF
--- a/lib/autoproj/cli/watch.rb
+++ b/lib/autoproj/cli/watch.rb
@@ -51,7 +51,7 @@ module Autoproj
             end
 
             def callback
-                notifier.stop
+                Thread.new { notifier.stop }
             end
 
             def create_file_watcher(file)


g-arjones:
Oh, hell.. :(

Shouldn't the thread be join()ed ?

doudou:
> Shouldn't the thread be join()ed ?

I thought I could forgo that. I can also make it without a thread, if you think that it looks too ugly.

g-arjones:
> I can also make it without a thread

How?

doudou:
> How?

[`#run`](https://github.com/guard/rb-inotify/pull/93/files#diff-b45361f26e8f95687c1c484833b7122fR225) is basically a `while !stop` around `#process`. So, just reimplement the while loop and have our own stop flag.

g-arjones:
Yeah.. I think I prefer the thread

doudou:
So ... good to go ?

g-arjones:
Yup :+1:

PR #242, opened 2019-02-21
allow packages to have internal dependencies

### LINKED ISSUE #240: initial import fails if python packages are used, but python is not installed
It looks like "python" is called by autobuild before the osdeps are installed

```
building initial autoproj import log, this may take a while
/usr/lib/ruby/2.3.0/open3.rb:199:in `spawn': No such file or directory - python (Errno::ENOENT)
	from /usr/lib/ruby/2.3.0/open3.rb:199:in `popen_run'
	from /usr/lib/ruby/2.3.0/open3.rb:95:in `popen3'
	from /root/.local/share/autoproj/gems/ruby/2.3.0/gems/autobuild-1.17.0/lib/autobuild/packages/python.rb:53:in `python_path'
	from /root/.local/share/autoproj/gems/ruby/2.3.0/gems/autobuild-1.17.0/lib/autobuild/packages/python.rb:89:in `update_environment'
```

g-arjones:
Fixes #240 

@doudou Thoughts?
@planthaber Could you please test if this fixes the issue you opened?

### DIFF
--- a/lib/autoproj/autobuild_extensions/dsl.rb
+++ b/lib/autoproj/autobuild_extensions/dsl.rb
@@ -164,8 +164,11 @@ def import_package(name, workspace: Autoproj.workspace, &block)
     package_common(:import, name, workspace: Autoproj.workspace, &block)
 end
 
-def python_package(name, workspace: Autoproj.workspace, &block)
-    package_common(:python, name, workspace: Autoproj.workspace, &block)
+def python_package(name, workspace: Autoproj.workspace)
+    package_common(:python, name, workspace: Autoproj.workspace) do |pkg|
+        pkg.internal_dependency 'python'
+        yield(pkg) if block_given?
+    end
 end
 
 def common_make_based_package_setup(pkg)
--- a/lib/autoproj/autobuild_extensions/package.rb
+++ b/lib/autoproj/autobuild_extensions/package.rb
@@ -3,7 +3,6 @@ module Autoproj
         module Package
             # Tags explicitely added with #add_tag
             attr_reader :added_tags
-
             attr_reader :optional_dependencies
 
             attr_reader :os_packages
@@ -24,6 +23,7 @@ module Autoproj
                 @optional_dependencies = Set.new
                 @description = PackageManifest.new(self, null: true)
                 @use_package_xml = false
+                @internal_dependencies = []
             end
 
             # Whether we should use a package.xml file present in this package
@@ -32,6 +32,10 @@ module Autoproj
             # @see use_package_xml=
             def use_package_xml?
                 @use_package_xml
+            end
+
+            def internal_dependencies
+                @internal_dependencies.dup
             end
 
             # Set {#use_package_xml?}
@@ -134,6 +138,21 @@ module Autoproj
                 dependencies.delete name
                 optional_dependencies.delete name
                 os_packages.delete name
+            end
+
+            # @api private
+            #
+            # Adds and 'internal' dependency to the package
+            # 'Internal' dependencies are actually autobuild dependencies
+            # for a given package type. It is assumed that only
+            # osdeps that do not rely on strict package managers will be
+            # declared as internal dependencies.
+            #
+            # The difference between a regular "os dependency" is that
+            # the internal dependency is installed right after the impport
+            # phase (before the actual osdeps phase)
+            def internal_dependency(osdeps_name)
+                @internal_dependencies << osdeps_name
             end
 
             def optional_dependency(name)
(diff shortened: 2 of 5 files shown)


planthaber:
Hi,

Yes, it fixes the issue.

Btw.: I saw that the import_package also has no dependencies yet. (none of wget,curl,tar,unzip) Bit i don't know what is really used.

https://github.com/rock-core/autoproj/blob/c8710396951e7a45c28b496e54fd7c525b459fc6/lib/autoproj/autobuild_extensions/dsl.rb#L163

g-arjones:
import_package is a "generic" package, the dependencies must be defined by the user in the package definition:

```ruby
import_package 'package' do |pkg|
    pkg.depends_on 'dep1'
end
```

planthaber:
But import_package uses the autobuild::Archive importer right?, which depends on tar,zip, etc?

But i think it won't be a similar issue, as tar should be installed by default is most distributions

g-arjones:
Ah, I get what you mean now.. This are actually importer dependencies, not package-type dependencies. You could, for example, have a `cmake_package` that depends on wget and tar if you have something like:

In an *.autobuild file:
```
cmake_package 'package'
```

And in a source.yml:

```yml
  - package:
    type: archive
    url: ftp://example.com/package.tar.gz
```

Similarly, you could have an `importer_package` that do not depend on either if the VCS definition type is just `git`.. Anyway, the archive importer does have its dependencies installed automatically by autoproj based on the vcs type: https://github.com/rock-core/autoproj/blob/ebc72e17e146d8083ee96060a8da9f2cfd86082c/lib/autoproj/ops/import.rb#L129 

The actual packages that will get installed are defined here: https://github.com/rock-core/autoproj/blob/ebc72e17e146d8083ee96060a8da9f2cfd86082c/lib/autoproj/default.osdeps#L131

Maybe we should add `wget` (on linux) and `curl (bsd)` there...

doudou:
> Maybe we should add `wget` (on linux) and `curl` (bsd) there...

No need, the archive importer now downloads using a pure ruby implementation.

lib/autoproj/autobuild_extensions/dsl.rb
```diff
@@ -164,8 +164,11 @@ def import_package(name, workspace: Autoproj.workspace, &block)
     package_common(:import, name, workspace: Autoproj.workspace, &block)
 end
 
-def python_package(name, workspace: Autoproj.workspace, &block)
-    package_common(:python, name, workspace: Autoproj.workspace, &block)
+def python_package(name, workspace: Autoproj.workspace)
+    package_common(:python, name, workspace: Autoproj.workspace) do |pkg|
+        pkg.internal_dependencies << 'python'
```
doudou: I'm not sold on the `internal_dependencies` name. These are basically dependencies required during import, right ? So what about `import_dependencies` or `early_dependencies` ? Something more descriptive of how they're different from the "normal" dependencies.

lib/autoproj/autobuild_extensions/package.rb
```diff
@@ -3,7 +3,7 @@ module AutobuildExtensions
         module Package
             # Tags explicitely added with #add_tag
             attr_reader :added_tags
-
+            attr_reader :internal_dependencies
```
doudou: I'm moving away from my very personal style of defining accessors for all instance variables. Even if you do add it, I'd prefer having an `add_` accessor for it.

g-arjones:
> the archive importer now downloads using a pure ruby implementation

Not through ftp...

doudou:
> Not through ftp...

True ...

lib/autoproj/autobuild_extensions/dsl.rb
g-arjones: > These are basically dependencies required during import, right?

Not really.. In this particular case, it required for settings up the environment for python source pacakges (so, after import). The general use case I guess is for dependencies of the `Autobuild::Packages` subclasses (autobuild's own dependencies, in a sense. which is why I suggested internal)

g-arjones: Also, I don't think it is a good idea to advertise this too much as there are caveats (we assume that these dependencies are always osdeps and that they do not depend on strict package managers). So, generally, the regular user should not use this

lib/autoproj/autobuild_extensions/package.rb
g-arjones: Sounds good to me...

lib/autoproj/autobuild_extensions/dsl.rb
doudou: Fair enough ... Good for `internal` then. Mark them explicitly as `@api private` then.

g-arjones:
Updated

doudou:
@planthaber could you test master now ? I'll do a bugfix release afterwards.

Issue #200, opened 2018-05-31
`autoproj watch` gets frozen if the user needs to answer a configuration question


doudou:
Ways to fix that:

- make `watch` fail in these cases, and wait for more changes. Should display a message that external tools can easily detect and show to the user (e.g. `ERROR: <msg>`)
- ?

g-arjones:
Shouldn't autoproj skip interactive configuration if subcommand == watch?

doudou:
Skipping the configuration is not really possible. The build setup depends on the configuration, so it can't be loaded or regen'ed.

This is why I was proposing to:
 - detect that interactive configuration is required
 - ignore the update event, and wait for more updates (i.e. user doing the configuration)

Unfortunately, the configuration will also be asked on e.g. update or build, which will have the same problem. Ultimately, we would have to be able to ask for the configuration value.

g-arjones:
Yes, you are right. I think warning that the build is not properly configured and waiting for another event is the way to go.

g-arjones:
Not an issue anymore, right? @doudou

doudou:
Still an issue, however:
1. I agree we could probably use `--no-interactive` to use a temporary configuration until the user does configure (by e.g. running update or build explicitly)
2. `--no-interactive` only works for flags that have a default

I guess that doing (1) would still be better than nothing, but (2) will still bite every now and then

PR #243, opened 2019-02-22
Permit autoproj to perform configuration noninteractively by relying …

2maz:
…on the default values

Avoid the need for updating seed configurations when new configurations options are introduced to autoproj.

### DIFF
--- a/lib/autoproj/cli/base.rb
+++ b/lib/autoproj/cli/base.rb
@@ -166,6 +166,7 @@ module Autoproj
             end
 
             def validate_options(args, options)
+                ws.config.interactive = options[:interactive]
                 self.class.validate_options(args, options)
             end
 
--- a/lib/autoproj/configuration.rb
+++ b/lib/autoproj/configuration.rb
@@ -22,6 +22,9 @@ module Autoproj
         attr_reader :displayed_options
         # The path to the underlying configuration file
         attr_reader :path
+        # Whether the configuration should be performed interactively
+        # or not
+        attr_accessor :interactive
 
         def initialize(path = nil)
             @config = Hash.new
@@ -43,16 +46,24 @@ module Autoproj
             @modified = false
         end
 
-        # Deletes the current value for an option
+        # Deletes the current configuration for all options or the one specified
         #
-        # The user will be asked for a new value next time the option is needed
+        # The user will be asked for a new value next time one of the reset
+        # options is needed
         #
-        # @param [String] the option name
+        # @param [String] name the option name (or by default nil to reset all
+        #   options)
         # @return the deleted value
-        def reset(name)
-            @modified ||= config.has_key?(name)
-            config.delete(name)
-            overrides.delete(name)
+        def reset(name = nil)
+            if name
+                @modified ||= config.has_key?(name)
+                config.delete(name)
+                overrides.delete(name)
+            else
+                @config.clear
+                @overrides.clear
+                @modified = true
+            end
         end
 
         # Sets a configuration option
@@ -177,10 +188,18 @@ module Autoproj
                 if current_value = config[option_name]
                     current_value = current_value.first
                 end
-                value = opt.ask(current_value)
+                is_default = false
+                if interactive?
+                    value = opt.ask(current_value, nil)
+                else
+                    value, is_default = opt.ensure_value(current_value)
+                    Autoproj.info "       using: #{value} (noninteractive mode)"
+                end
                 @modified = true
-                config[option_name] = [value, true]
-                displayed_options[option_name] = value
+                if !is_default
+                    config[option_name] = [value, true]
+                    displayed_options[option_name] = value
+                end
                 value
             else
                 raise ConfigError.new, "undeclared option '#{option_name}'"
@@ -440,6 +459,21 @@ module Autoproj
             set('randomize_layout', value, true)
         end
 
+        # Returns true if the configuration should be performed interactively
+        #
+        # @return [Boolean]
+        # @see interactive=
+        def interactive?
+            if !interactive.nil?
+                return interactive
+            elsif ENV['AUTOPROJ_NONINTERACTIVE'] == '1'
+                return false
+            elsif has_value_for?("interactive")
+                return get('interactive')
+            end
+            true
+        end
+
         DEFAULT_UTILITY_SETUP = Hash[
             'doc' => true,
             'test' => false]
@@ -545,7 +579,11 @@ module Autoproj
         def to_hash
             result = Hash.new
             @config.each do |key, (value, _)|
-                result[key] = value
+                if declared_options.include?(key)
+                    result[key] = declared_options[key].ensure_value(value)
+                else
+                    result[key] = value
+                end
             end
             overrides.each do |key, value|
                 result[key] = value
--- a/lib/autoproj/workspace.rb
+++ b/lib/autoproj/workspace.rb
@@ -236,10 +236,16 @@ module Autoproj
             File.join(config_dir, OVERRIDES_DIR)
         end
 
+        # Load the configuration for this workspace from
+        # config_file_path
+        #
+        # @param [Boolean] reset Set to true to replace the configuration object,
+        #   set to false to load into the existing
+        # @return [Configuration] configuration object
         def load_config(reconfigure = false)
-            @config = Configuration.new(config_file_path)
             if File.file?(config_file_path)
-                config.load(reconfigure: reconfigure)
+                config.reset
+                config.load(path: config_file_path, reconfigure: reconfigure)
                 if raw_vcs = config.get('manifest_source', nil)
                     manifest.vcs = VCSDefinition.from_raw(raw_vcs)
                 else
--- a/test/cli/test_reconfigure.rb
+++ b/test/cli/test_reconfigure.rb
@@ -1,0 +1,42 @@
+require 'autoproj/test'
+require 'autoproj/cli/main'
+require 'autoproj/cli/reconfigure'
+require 'timeout'
+
+module Autoproj
+    module CLI
+        describe Reconfigure do
+            attr_reader :ws
+            before do
+                option_name = "custom-configuration-option"
+                default_value = "option-defaultvalue"
+
+                @ws = ws_create(make_tmpdir, partial_config: true)
+            end
+            after do
+                Autoproj.verbose = false
+            end
+            describe "#reconfigure" do
+                def run_command(*args)
+                    capture_subprocess_io do
+                        ENV['AUTOPROJ_CURRENT_ROOT'] = ws.root_path.to_s
+                        in_ws do
+                            Main.start([*args,"--debug"], debug: true)
+                        end
+                    end
+                end
+
+                it "reconfigure should run interactively" do
+                    assert_raises Timeout::Error do
+                        Timeout.timeout(3) do
+                            run_command 'reconfigure'
+                        end
+                    end
+                end
+                it "reconfigure should run non interactively" do
+                    run_command 'reconfigure','--no-interactive'
+                end
+            end
+        end
+    end
+end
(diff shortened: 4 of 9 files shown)


planthaber:
+1

Good change, makes maintaining CI way easier.

g-arjones:
In principle, I like the idea... I wonder if we should maybe skip saving the configuration if `AUTOPROJ_NONINTERACTIVE` is enabled and also maybe add a `--no-interactive` option in the CLI.

What do you guys think?

2maz:
> skip saving the configuration if `AUTOPROJ_NONINTERACTIVE` is enabled

what would be the use case, seed-configurations might still be used in combination with a non interactive mode

> maybe add a `--no-interactive` option in the CLI.

+1  but maintaining the control through the env var

g-arjones:
> what would be the use case

Automation tools and/or the vscode rock/autoproj extension during operations other than install/bootstrap. Also, this could be a fix/workaround for #200 but we would still want the user to be asked about the configuration in the future

> but maintaining the control through the env var

Sounds good..

planthaber:
What do you think about to also set --no-color and --no-progress when AUTOPROJ_NONINTERACTIVE is set?

2maz:
> Automation tools and/or the vscode rock/autoproj extension during operations other than install/bootstrap. Also, this could be a fix/workaround for #200 but we would still want the user to be asked about the configuration in the future

Sorry, still not clear to me as to why you do not want to save the configuration. 
From my (most likely limited) point of view, not to save the config means: you always want to query the anserstick to the defaults?
However, with non-interactive mode (for CI, unattended deployments ..) a user might still provide a seed-configuration to be used for the bootstrap. 
This configuration should be applied and it should stick - independently of the changes to the defaults might be made.

doudou:
> This configuration should be applied and it should stick - independently of the changes to the defaults might be made.

This actually sounds weird. If non-interactive picks up defaults, then why don't you want to use the up-to-date defaults, in case they change ? Especially in CI, it would make sure that the default build actually builds...

I actually like @g-arjones idea of not saving the values (if it's reasonably easy to do), allowing to bootstrap non-interactively and then do a manual run to check the configuration(s). This seems compatible with having a seed file (which only consists of reading values, not writing them).

Finally, I would very much prefer that the flag is made available on the configuration object (`Configuration#interactive?`). I'd be fine to have it only through the envvar for now, but in the future, I'd also prefer being allowed to pass it explicitly on the command line and/or through the seed file (e.g. during bootstrap, the way `--separate-prefixes` works). Putting it on the configuration object would allow for this. Given that `BuildOption` is stored on a configuration object in the first place, this seems reasonably easy to do (have `#interactive?` on the object, and pass it as argument to `#ask` instead of implicitly through the env var).

Also, **please write tests**.

doudou:
> What do you think about to also set --no-color and --no-progress when AUTOPROJ_NONINTERACTIVE is set?

:-1: You can run noninteractively in a terminal emulator that supports color and/or progress.

2maz:
> 
> This actually sounds weird. If non-interactive picks up defaults, then why don't you want to use the up-to-date defaults, in case they change ? Especially in CI, it would make sure that the default build actually builds...

Well I did not say that. To be clear I am stressing the fact that the configuration options which a user sets explicitly via the seed-configuration should be maintained.
These should not change just because a default options has changed.
That is why have been asking for clarification on the issue of not saving "the configuration".
if @g-arjones is suggesting (and @doudou s response now implies this) to skip saving only those configuration options that are set through a fallback to default - then I am fully with you.

> then do a manual run to check the configuration(s). 

What is that supposed to mean?

> Finally, I would very much prefer that the flag is made available on the configuration object (`Configuration#interactive?`). I'd be fine to have it only through the envvar for now, but in the future, I'd also prefer being allowed to pass it explicitly on the command line and/or through the seed file (e.g. during bootstrap, the way `--separate-prefixes` works). Putting it on the configuration object would allow for this. Given that `BuildOption` is stored on a configuration object in the first place, this seems reasonably easy to do (have `#interactive?` on the object, and pass it as argument to `#ask` instead of implicitly through the env var).

Sure, I can put something along these lines together

g-arjones:
Exactly what I meant, sorry for not being clear enough. I guess we are all on the same page then..

doudou:
> What is that supposed to mean?

Well ... start a non-interactive bootstrap/update/build and go to lunch. When you come back, run `amake` interactively, this time, and check the flags. You only need to rebuild stuff you would have changed.

test/cli/test_reconfigure.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    module CLI
+        describe Reconfigure do
+            attr_reader :ws
+            before do
+                option_name = "custom-configuration-option"
+                default_value = "option-defaultvalue"
+
+                @ws = ws_create(make_tmpdir, partial_config: true)
+                # @doudou: What am I doing wrong here,
```
2maz: @doudou: I seem to be missing something here to correctly initiate the configuration setup, so suggestions welcome - I would like to get rid of the above partial_config workaround . ..

2maz:
Update: Implemented something that hopefully covers all mentioned requirements. I need some help for the test side, however, see comment above.

test/cli/test_reconfigure.rb
g-arjones: Is it really worth to spawn a separate process for this? Why don't you just call `Configuration#reconfigure!` and verify that `BuildOption#ask` is called for a declared config?

2maz: > Is it really worth to spawn a separate process for this?

It is a test, why should another process pose a problem? 

> Why don't you just call Configuration#reconfigure! and verify that BuildOption#ask is called for a declared config?

Wouldn't that be testing only the Configuration object, while here I want to cover the CLI?

lib/autoproj/configuration.rb
```diff
@@ -440,6 +445,27 @@ def randomize_layout=(value)
...
+        def interactive=(flag)
+            set('interactive', flag, true)
+        end
+
+        # Returns true if the configuration should be performed interactively
+        #
+        # @return [Boolean]
+        # @see interactive=
+        def interactive?
+            if ENV['AUTOPROJ_NONINTERACTIVE'] == '1' || !Autoproj.interactive
```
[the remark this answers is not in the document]
2maz: Good point. Though I have to change the default on Autoproj.interactive then to nil, in order to identify a CLI based-setting.

2maz:
> I don't see how you are skipping saving options that had defaults chosen due to non-interactive mode...

sorry, apparently missed that one ..

2maz:
> I don't see how you are skipping saving options that had defaults chosen due to non-interactive mode...

Added the feature ...

lib/autoproj/configuration.rb
```diff
@@ -5,6 +5,11 @@
 require 'backports/2.4.0/true_class/dup'
 
 module Autoproj
+    class << self
+        attr_accessor :interactive
+    end
+    @interactive = nil
```
g-arjones: Initializing an instance variable with `nil` is not necessary

lib/autoproj/cli/base.rb
```diff
@@ -176,11 +176,13 @@ def self.validate_options(args, options)
                     debug: false,
                     color: TTY::Color.color?,
                     progress: TTY::Color.color?,
-                    parallel: nil
+                    parallel: nil,
+                    interactive: true
```
g-arjones: Shouldn't this be `nil`?

test/cli/test_reconfigure.rb
g-arjones: The other process does not pose a problem it just makes testing harder (or impossible). You end up with stuff like:

```ruby
it "reconfigure should run non interactively" do
    run_command 'reconfigure','--no-interactive'
end
```

Which is not really a UNIT test...

g-arjones:
Other than a few style issues (@doudou will probably have something to say about those) I think we are good to go

lib/autoproj/cli/base.rb
2maz: Right and I have to change the default in the cli/main.rb as well.

lib/autoproj/cli/base.rb
```diff
@@ -176,11 +176,13 @@ def self.validate_options(args, options)
...
                     color: TTY::Color.color?,
                     progress: TTY::Color.color?,
-                    parallel: nil
+                    parallel: nil,
+                    interactive: nil
 
                 Autoproj.silent = options[:silent]
                 Autobuild.color = options[:color]
                 Autobuild.progress_display_enabled = options[:progress]
+                Autoproj.interactive = options[:interactive]
```
doudou: The only purpose of this is to pass the CLI argument to the config object. Just set the argument (if given) to the config object instead, removing the middle-man.

All these global flags have their history ... but they're meant to disappear slowly ...

lib/autoproj/configuration.rb
```diff
@@ -177,7 +184,10 @@ def configure(option_name)
                 if current_value = config[option_name]
                     current_value = current_value.first
                 end
-                value = opt.ask(current_value)
+                value = opt.ask(current_value, nil, interactive: interactive?)
```
doudou: It's weird to **ask** anything if it's not meant to actually **be asked**.

Why not leave `#ask` alone, and simply return the default value, not even storing it in the config hash ? That would remove the need for the `@use_default` variable. You would however need to take this into account in `#to_hash` (filling the defaults from options that haven't been asked yet)

lib/autoproj/cli/base.rb
2maz: Shifted the flag into the Autoproj::Configuration namespace if that is what you intended. Using an instance variable fails, since the Autoproj.config object seems to be replaced at a later stage.

lib/autoproj/configuration.rb
2maz: Just followed your initial suggestion, but it should be less 'weird' now

lib/autoproj/cli/base.rb
doudou: `Autoproj.config` is really only meant for the autobuild files. A workspace config is available through the config object (`ws.config` in CLI classes). If it doesn't work, I can help and debug it. The bottom line being that I'd rather not have yet-another global added.

lib/autoproj/workspace.rb
```diff
@@ -236,8 +236,19 @@ def overrides_dir
...
 
-        def load_config(reconfigure = false)
-            @config = Configuration.new(config_file_path)
+        # Load the configuration for this workspace from
+        # config_file_path
+        #
+        # @param [Boolean] reset Set to true to replace the configuration object,
+        #   set to false to load into the existing
+        # @return [Configuration] configuration object
+        def load_config(reconfigure = false, replace: true)
```
doudou: Adding this optional behavior is going to be very fragile. From what I see, the "resetting the configuration" was there because everything in Configuration, so far, was saved to disk, and it was therefore easier.

Since it's not the case anymore, I'd suggest to simply change Configuration to make the reset unnecessary, for instance:

~~~
def reset
   @config.clear
   @modified = true
end
~~~

and then just change `load_config` to do a reset + load.

2maz: Ack, that's the better approach

2maz: Updated accordingly

doudou: Great ! Thanks !

It seems that it broke a test in `test_workspaces.rb` though ... :P

2maz: Assumptions of the test for flexmock on creating a new configuration were not true anymore, so updated the test.

doudou:
Thanks for the work Thomas !

2maz:
got a bit meatier than I thought initially, but happy to had your reviews @doudou @g-arjones

g-arjones:
@2maz 👍 

any time!

PR #245, opened 2019-02-28
Add "autoproj version"

2maz:


lib/autoproj/cli/version.rb
```diff
@@ -0,0 +1,18 @@
+require 'autoproj/cli/base'
+module Autoproj
+module CLI
+    class Version < Base
+        def run(args, options = Hash.new)
+            puts "autoproj version: #{Autoproj::VERSION}"
+            if options[:include_deps]
+                msg, status = Open3.capture2e("gem dependency autoproj -v #{Autoproj::VERSION}")
```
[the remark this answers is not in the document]
2maz: Updated to list requirements and currently active gems

lib/autoproj/cli/version.rb
```diff
@@ -0,0 +1,62 @@
...
+
+module Autoproj
+module CLI
+    class Version < Base
+        # List the version of autoproj and optionally include
+        # the installed dependencies with information about the requirement
+        # and the actual used version
+        def run(args, options = Hash.new)
+            puts "autoproj version: #{Autoproj::VERSION}"
+            if options[:deps]
```
doudou: Please return early, that is

~~~
return unless options[:dep]
~~~

lib/autoproj/cli/version.rb
```diff
@@ -0,0 +1,62 @@
...
+                dependency = Gem::Deprecate.skip_during do
+                    Gem::Dependency.new "autoproj", Autoproj::VERSION
+                end
+                autoproj_spec = dependency.matching_specs
+                if !autoproj_spec.empty?
```
doudou: Same comment about returning early.

lib/autoproj/cli/version.rb
```diff
@@ -0,0 +1,62 @@
...
+                    autoproj_spec.first.dependencies.each do |dep|
+                        puts "        #{dep}: #{installed_deps[dep.name] || 'n/a'}"
+                        installed_deps.delete(dep.name)
+                    end
+                    puts "    implicit dependencies:"
+                    installed_deps.keys.sort.each do |name|
+                        puts "        #{name}: #{installed_deps[name]}" unless name == "autoproj"
+                    end
+                else
+                    Autoproj.warn "Dependencies and their versions could not be retrieved -- is gem packager installed?"
```
doudou: I don't think this is the right warning message anymore ... At this point, I would classify the error as "internal error". Autoproj is always bootstrapped using bundler/gem.

lib/autoproj/cli/version.rb
```diff
@@ -0,0 +1,62 @@
...
+        #
+        # @return [Array<Gem::Dependency>] all known dependencies
+        def collect_dependencies(dependency, known_dependencies: [])
+            dep_spec = dependency.matching_specs
+            if !dep_spec.empty?
+                dep_spec = dep_spec.first
+                known_dependencies << dep_spec
+                if !dep_spec.dependencies.empty?
+                    dep_spec.dependencies.each do |dep|
+                        if !known_dependencies.any? { |d| d.name == dep.name }
```
doudou: Looks like [Gem::Dependency](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/dependency.rb) is compatible with `==` and more generally Hash/Set. Turn `known_dependencies` into a Set and have a much nicer test (and remove the need for `uniq` below)

lib/autoproj/cli/version.rb
```diff
@@ -0,0 +1,62 @@
...
+        # @param [Array<Gem::Dependency] list of already known dependencies
+        #
+        # @return [Array<Gem::Dependency>] all known dependencies
+        def collect_dependencies(dependency, known_dependencies: [])
+            dep_spec = dependency.matching_specs
+            if !dep_spec.empty?
+                dep_spec = dep_spec.first
+                known_dependencies << dep_spec
+                if !dep_spec.dependencies.empty?
+                    dep_spec.dependencies.each do |dep|
```
doudou: `.each` on an empty collection just does nothing, you don't need the if to guard it.

lib/autoproj/cli/version.rb
```diff
@@ -0,0 +1,62 @@
...
+                        if !known_dependencies.any? { |d| d.name == dep.name }
+                            dependencies = collect_dependencies(dep, known_dependencies: known_dependencies)
+                            known_dependencies.unshift(*dependencies)
```
doudou: `.concat(dependencies)` ? (`.merge` for a Set)

lib/autoproj/cli/version.rb
```diff
@@ -0,0 +1,62 @@
...
+            if options[:deps]
+                dependency = Gem::Deprecate.skip_during do
+                    Gem::Dependency.new "autoproj", Autoproj::VERSION
+                end
+                autoproj_spec = dependency.matching_specs
+                if !autoproj_spec.empty?
+                    known_dependencies = collect_dependencies(dependency)
+                    installed_deps = {}
+                    known_dependencies.flatten.uniq do |spec|
+                        installed_deps[spec.name] = spec.version
```
doudou: If that's your ultimate goal, why not make `collect_dependencies` return that (which would be a good alternative to the Array->Set change I suggest below)

2maz: Updated along  these lines

lib/autoproj/cli/version.rb
doudou: Ping comment above

lib/autoproj/cli/version.rb
```diff
@@ -0,0 +1,53 @@
...
+        def collect_dependencies(dependency, known_dependencies: {})
+            dep_spec = dependency.matching_specs
+            return known_dependencies if dep_spec.empty?
+
+            dep_spec = dep_spec.first
+            known_dependencies[dep_spec.name] = dep_spec.version
+            dep_spec.dependencies.each do |dep|
+                if !known_dependencies.has_key?(dep.name)
+                    dependencies = collect_dependencies(dep, known_dependencies: known_dependencies)
+                    known_dependencies.merge(dependencies)
```
doudou: This does nothing (and doesn't need to, interestingly, since you pass the known_dependencies hash to the callee).

lib/autoproj/cli/version.rb
2maz: sorry I missed that one

lib/autoproj/cli/version.rb
2maz: updated

doudou:
Merged. I just realized that in a CI context, you'd best just cat `.autoproj/Gemfile.lock`, which is more complete (it for instance would show if you were using autoproj from git, were you using unreleased functionality of autoproj the way we're currently doing in our CI)

2maz:
Thx, good to know.

PR #246, opened 2019-03-01
added information to versions result

sbolivarm:


lib/autoproj/ops/snapshot.rb
```diff
@@ -140,6 +140,10 @@ def snapshot_packages(packages, target_dir = nil, only_local: true)
                         next
                     end
 
+
+                vcs_info['fingerprint'] = package.autobuild.fingerprint
+                vcs_info['dependencies'] = package.autobuild.dependencies
```
doudou: Dependencies ... why ? It really isn't its place. The snapshot data is meant as importer information. Even `fingerprint` seems to be out of place somewhat here (but it feels less wrong than `dependencies`).

The 'versions' is actually meant as importer information - it really is an override file - and both fingerprint and dependencies feel out of place here. I would suggest that we do a benchmark on a realistic checkout (after the performance fix in the autobuild PR) and see whether it would make sense to generate the fingerprint unconditionally in the installation manifest.

lib/autoproj/ops/snapshot.rb
```diff
@@ -140,6 +140,10 @@ def snapshot_packages(packages, target_dir = nil, only_local: true)
                         next
                     end
 
+
```
doudou: Stray empty line

lib/autoproj/ops/snapshot.rb
sbolivarm: Dependencies should definitely not be there, I was initially using it for debugging purposes and I forgot to remove it.

lib/autoproj/ops/snapshot.rb
sbolivarm: removed

lib/autoproj/ops/snapshot.rb
```diff
@@ -140,6 +140,8 @@ def snapshot_packages(packages, target_dir = nil, only_local: true)
                         next
                     end
 
+                vcs_info['fingerprint'] = package.autobuild.fingerprint
```
g-arjones: This is weird. If the importer supports fingerprinting this should be returned by `#snapshot` from the importer subclass...

doudou: In this case, `#fingerprint` is calculated by the package (because calculated recursively), not the importer.

In any case, we're discussing moving this to the install importer. We'll do some performance measurement first, just to be sure it's not too expensive.

doudou:
@sbolivarm: I've tested the performance to see whether we could put it in the installation manifest.... and we cannot. It basically doubles the duration of `envsh` because of the cost of shelling out to git.

We will maybe one day migrate to a libgit wrapper ([this one](https://github.com/libgit2/rugged) looks nice), but not today. So, let's leave it in the `versions` output. I would however prefer that you save it optionally (under a `--fingerprint` option)

g-arjones:
Ok, the intention is clear now. Though I like the idea, I am still unsure about the implementation. Unfortunately I can't suggest anything concrete at this point. I will, however, highlight a few points in case you guys have mixed feelings about this as well:

1. The goal of `versions` is to provide additional configuration for the importers so that the current state of the source tree can be more easily replicated. The fact that you have to "arbitrarily" inject information in the hash (that is named `vcs_info`, btw) is an evidence of that;

2. I don't think it should be allowed having a package/importer that does not support fingerprinting and it's definitely not safe to assume that a package that has a dependency on a package that does not support fingerprinting is up-to-date. Personally, I would rather either raise an error or show a big fat warning and assume that the package should be rebuilt anyway;

3. This has a very specific use case and though it must be something that has to be reliable, it is not cool to annoy users that are not using the feature neither with a warning nor by outright breaking the build (possibly because the workspace has something that does not support fingerprinting). Personally, I would prefer having a dedicated subcommand in the CLI for this or maybe optionally adding the fingerprint to the installation-manifest. It would also help improving (1);

4.  Current implementation overlooks the fact that upgrading osdeps may also break the build, would be cool if we could rely on CI to avoid that. We could use the package managers during fingerprint computation to get osdeps versions during this process;

I hope I could provide any useful insights...

doudou:
> 1. is an evidence of that;

I'm not sure an evidence of what it is ... If you mean that it's weird to arbitrarily inject information in the `vcs_info` hash, I fully agree. This is why I've commented to calculate the fingerprint only with a specific command-line argument.

>  2.

I think we can't assume that all importers will support fingerprinting. I agree that recursive fingerprinting should fail (return nil) if any of the dependencies can't compute its fingerprint, and that the cache handling should handle this gracefully. @sbolivarm ?

> 3.

Solved by the specific command-line argument. Unfortunately, the current git importer implementation shells out, which makes the feature too expensive to include in the installation manifest (on a typical workspace I have here, it costs ~2.5s more to do envsh)

> 4. 

We elected to ignore this for now. The CI will usually have a stable binary osdep environment (i.e. APT stuff) and one can make sure that caches are not shared across build environments if there is more than one. More complicated for the Gem/Pip stuff since we don't build the dependent packages, but we would want to run tests. As I said, for now, we thought it would already be a good approximation, which would make building PRs practical - something that IMO will have a much more positive impact than taking into account the osdeps.

lib/autoproj/cli/build.rb
```diff
@@ -42,6 +43,16 @@ def run(selected_packages, options)
                     pkg.disable
                 end
 
+                # and disable packages listed in the not command if there are any
+                if build_options[:not]
+                    source_packages.each do |pkg_name|
```
doudou: I'd prefer that you integrate this with the loop above, which have the same purpose (disable what's not to be built).

Moreover, don't check for nil in `not`. Make sure that `build_options[:not]` is always set instead.

shell/completion/amake_bash
```diff
@@ -61,6 +61,7 @@ __amake() {
         --no-tool
         --confirm
         --no-confirm
+        --not
```
doudou: Good catch ... I didn't thought about updating the autocompletion ...

lib/autoproj/cli/build.rb
sbolivarm: Is there any reason to use `next` in the disabling loop? I was thinking to do something like this:
```
if  source_packages.include?(pkg.name) || 
   (build_options[:not] && !build_options[:not].include?(pkg.name))
            Autobuild.warn 'disabling ' + pkg.name
            pkg.disable
 end
```

sbolivarm: I have refactored the loops with a single one using an active_packages variable that contains source packages minus packages in the not option if there are any.

PR #252, opened 2019-03-27
add auto-detection of autotools packages

doudou:


lib/autoproj/autobuild_extensions/dsl.rb
```diff
@@ -95,29 +95,39 @@ def self.find_topmost_directory_containing(dir, glob_pattern = nil)
 
     # Tries to find a handler automatically for 'full_path'
     def self.package_handler_for(full_path)
+        pyglob = File.join(File.basename(full_path), "*.py")
         if !Dir.enum_for(:glob, File.join(full_path, "*.orogen")).to_a.empty?
-            return "orogen_package", full_path
+            ["orogen_package", full_path]
+        elsif File.file?(File.join(full_path, "Makefile.am"))
```
g-arjones: Given the fact that `Autobuild::Autotools` is able to detect needed stages I would implement this like:

1. The package has a pre-generated `configure` script;
2. If the package DOES NOT have a pre-generated `configure` script, it MUST have either a `configure.ac` or `configure.in` file

Everything else should be detected by the package handler... Packages that do not use automake (so no Makefile.am) are somewhat common. Also, the way I see it, it is totally legit to use autoconf just to generate a `config.h` with detected system features (in which case there would be no Makefile.in as well).

I believe the package handler already deals with this kind of setup just fine.

doudou: I agree. The problem is that in the pre-cmake world, there are quite a few packages with a 'configure' script that are not autotools-based.

Maybe just checking configure.ac/configure.in ? I agree that there's quite a few packages with autoconf and not automake, but I'm not aware of the other way around.

g-arjones: Yep.. Makes sense

doudou: Done. I kept the Makefile.am logic, as it allows to detect a package from a subdirectory (were one to do `amake` in a subdirectory).

g-arjones:
Looks good... Btw, did you notice `autoproj test exec` not showing progress anymore? 

Got this report from people here just now..

doudou:
Didn't notice, no. I'm unfortunately not running `autoproj test` these days.

g-arjones:
Removing rock-core/autobuild#67 seems to fix the issue somehow..

doudou:
> Removing [rock-core/autobuild#67](https://github.com/rock-core/autobuild/pull/67) seems to fix the issue somehow..

In autobuild ? Weird. Well ... not completely surprising either unfortunately. These huge refactorings always have some weird side effects when the test suite is incomplete...

g-arjones:
Yep.. Workaround  for me for now is running with either `--debug` or `--tool`

PR #253, opened 2019-03-27
cli: fix error message on load error in 'build'

doudou:
The error message would be related to the failure of export_env_sh,
not the actual load error. Just don't export env.sh if loading failed.


g-arjones:
Looks good to me.. Did not run the tests though

PR #256, opened 2019-04-14
Cli test validation

g-arjones:


doudou:
Could you maybe write a test ? That would be great.

g-arjones:
Definitely, but there are no tests for the `test` subcommand so it will take a while..

doudou:
> Definitely, but there are no tests for the `test` subcommand so it will take a while..

From my perspective, just add tests that check the specific regression you spotted. We can extend to cover the rest later.

There start to be quite a few examples in `test/cli/` to build from

g-arjones:
Sounds good to me 👍

g-arjones:
Here you go... Would you please release this along with the other changes currently in master (autobuild as well). Appreciate that...

doudou:
> Would you please release this along with the other changes currently in master (autobuild as well). Appreciate that...

I'm trying to get the test suite to pass before I release everything. Have had little time to invest, unfortunately.

PR #260, opened 2019-05-21
pin xdg gem version

g-arjones:
xdg 2.2.4 is broken: https://github.com/rubyworks/xdg/issues/7


doudou:
Given that the PR says "Be aware major breaking changes will be introduced in the upcoming 3.0.0 release (2019-06-01).", maybe we should indeed pin to 2.5.0. What do you think ?

g-arjones:
Sounds good to me (assuming you meant 2.2.5)...

PR #268, opened 2019-08-19
allow managing doc dependencies through cli

g-arjones:
@doudou How would feel about making doc_depend off by default (to be consistent with test_depend)?


doudou:
I always try to avoid changing default behavior (thanks, btw, for keeping `autoproj doc` the same).

A.k.a. I'd rather keep them on by default, especially since there's in the end few of them.

PR #272, opened 2019-08-29
bundler: allow setting build flags from Autoproj configuration

doudou:
Bundler allows configuring per-gem build flags, to let the gem
that have C extensions find their dependencies.

This commit adds the ability to set these flags within an
Autoproj workspace, from e.g. within overrides.rb:

~~~ruby
Autoproj::PackageManagers::BundlerManager
.configure_build_for(
  Autoproj.workspace,
  'gemname', '--some=other')
~~~

### DIFF
--- a/lib/autoproj/package_managers/bundler_manager.rb
+++ b/lib/autoproj/package_managers/bundler_manager.rb
@@ -88,8 +88,79 @@ module Autoproj
                     end
                 end
 
-                if bundle_rubylib = discover_bundle_rubylib(silent_errors: true)
+                apply_build_config(prefix_gems)
+
+                if (bundle_rubylib = discover_bundle_rubylib(silent_errors: true))
                     update_env_rubylib(bundle_rubylib, system_rubylib)
+                end
+            end
+
+            # Enumerate the per-gem build configurations
+            def per_gem_build_config
+                ws.config.get('bundler.build', {})
+            end
+
+            # Set the build configuration for the given gem
+            #
+            # This is meant to be used from the Autoproj configuration files,
+            # e.g. overrides.rb or package configuration
+            def self.configure_build_for(gem_name, build_config, ws: Autoproj.workspace)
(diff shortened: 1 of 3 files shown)


lib/autoproj/package_managers/bundler_manager.rb
```diff
@@ -88,11 +88,82 @@ def initialize_environment
...
+            # Enumerate the per-gem build configurations
+            def per_gem_build_config
+                ws.config.get('bundler.build', {})
+            end
+
+            # Set the build configuration for the given gem
+            #
+            # This is meant to be used from the Autoproj configuration files,
+            # e.g. overrides.rb or package configuration
+            def self.configure_build_for(ws, gem_name, build_config)
```
g-arjones: Could we move `ws` to last and have a default to `Autoproj.workspace`?

lib/autoproj/package_managers/bundler_manager.rb
```diff
@@ -88,11 +88,82 @@ def initialize_environment
...
+                c = ws.config.get('bundler.build', {})
+                c[gem_name] = build_config
+                ws.config.set('bundler.build', c)
+            end
+
+            # Removes build configuration flags for the given gem
+            #
+            # This is meant to be used from the Autoproj configuration files,
+            # e.g. overrides.rb or package configuration
+            def self.remove_build_configuration_for(ws, gem_name)
```
g-arjones: Same here

lib/autoproj/package_managers/bundler_manager.rb
doudou: Makes a lot of sense. :+1:

doudou: Done

lib/autoproj/package_managers/bundler_manager.rb
doudou: Done

PR #273, opened 2019-08-30
generate a build report file on update as well

doudou:
The build report was created only on build so far, which means that
nothing would be reported if update fails (!)

Generate one on update too.

### DIFF
--- a/lib/autoproj/ops/import.rb
+++ b/lib/autoproj/ops/import.rb
@@ -12,9 +12,10 @@ module Autoproj
             end
             attr_writer :auto_exclude
 
-            def initialize(ws)
+            def initialize(ws, report_path: nil)
                 @ws = ws
                 @auto_exclude = false
+                @report_path = report_path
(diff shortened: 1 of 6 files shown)


lib/autoproj/ops/import.rb
```diff
@@ -477,6 +483,33 @@ def process_post_import_blocks(pkg)
...
+                        build_invoked: pkg.build_invoked?,
+                        failed: pkg.failed?,
+                        imported: pkg.imported?,
+                        prepared: pkg.prepared?,
+                        built: pkg.built?
+                    }
+                end
+
+                build_report = JSON.pretty_generate({
+                    build_report: {
```
g-arjones: build_report or update_report ?

doudou: The latter... thanks

doudou: Updated

PR #276, opened 2019-09-02
bundler: apply per-gem configuration right before running install

doudou:
Associated build test: https://github.com/rock-core/rock.build-tests-buildconf/pull/3

Doing it in initialize_environment is actually pretty limited:
the only place where code gets run before initialize_environment
is autoproj/init.rb

Move updating the bundler config right before it is needed,
that is in run_bundler_install

### DIFF
--- a/lib/autoproj/package_managers/bundler_manager.rb
+++ b/lib/autoproj/package_managers/bundler_manager.rb
@@ -88,15 +88,13 @@ module Autoproj
                     end
                 end
 
-                apply_build_config(prefix_gems)
-
                 if (bundle_rubylib = discover_bundle_rubylib(silent_errors: true))
                     update_env_rubylib(bundle_rubylib, system_rubylib)
                 end
             end
 
             # Enumerate the per-gem build configurations
-            def per_gem_build_config
+            def self.per_gem_build_config(ws)
                 ws.config.get('bundler.build', {})
             end
 
@@ -123,11 +121,12 @@ module Autoproj
             # @api private
             #
             # Apply configured per-gem build configuration options
-            # 
-            # @param [String] root_dir the path of the Bundler workspace root
-            #   directory. In Autoproj workspaces, this is {prefix_dir}/gems by
-            #   default
-            def apply_build_config(root_dir)
+            #
+            # @param [Workspace] ws the workspace whose bundler configuration
+            #   should be updated
+            # @return [void]
+            def self.apply_build_config(ws)
+                root_dir = File.join(ws.prefix_dir, 'gems')
                 current_config_path = File.join(root_dir, ".bundle", "config")
                 current_config =
                     if File.file?(current_config_path)
@@ -137,7 +136,7 @@ module Autoproj
                     end
 
                 build_config = {}
-                per_gem_build_config.each do |name, conf|
+                per_gem_build_config(ws).each do |name, conf|
                     build_config[name.upcase] = conf
                 end
 
@@ -253,6 +252,8 @@ module Autoproj
                 if binstubs
                     options << "--binstubs" << binstubs
                 end
+
+                apply_build_config(ws)
 
                 connections = Set.new
                 run_bundler(ws, 'install', *options, gem_home: gem_home, gemfile: gemfile) do |line|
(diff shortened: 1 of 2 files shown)


lib/autoproj/package_managers/bundler_manager.rb
```diff
@@ -127,7 +135,8 @@ def self.remove_build_configuration_for(gem_name, ws: Autoproj.workspace)
             # @param [String] root_dir the path of the Bundler workspace root
             #   directory. In Autoproj workspaces, this is {prefix_dir}/gems by
             #   default
-            def apply_build_config(root_dir)
+            def self.apply_build_config(ws)
```
g-arjones: Default to Autoproj.workspace?

lib/autoproj/package_managers/bundler_manager.rb
```diff
@@ -127,7 +135,8 @@ def self.remove_build_configuration_for(gem_name, ws: Autoproj.workspace)
             # @param [String] root_dir the path of the Bundler workspace root
```
g-arjones: Update documentation

lib/autoproj/package_managers/bundler_manager.rb
```diff
@@ -88,18 +88,26 @@ def initialize_environment
...
-                apply_build_config(prefix_gems)
-
                 if (bundle_rubylib = discover_bundle_rubylib(silent_errors: true))
                     update_env_rubylib(bundle_rubylib, system_rubylib)
                 end
             end
 
             # Enumerate the per-gem build configurations
-            def per_gem_build_config
+            def self.per_gem_build_config(ws)
```
g-arjones: Same, I would prefer to have Autoproj.workspace as  the default argument

lib/autoproj/package_managers/bundler_manager.rb
doudou: SInce this is not meant to be used from within the autobuild files, I wanted to avoid having a reference to Autoproj.workspace, which is set only there.

What do you think ?

doudou: Since it this method (as well as per_gem_build_config below) is not meant to be used from with autobuild files, I wanted to avoid having a reference to Autoproj.workspace, which is used only there.

What do you think ?

g-arjones: I just don't find an use case where I would want to use these methods without `Autoproj.workspace` as the argument (regardless of whether I am working in autobuild files or otherwise), having it default to `Autoproj.workspace` simplifies the API, IMO. In fact, the only reason for passing these as arguments is to making testing easier. Do you see any other reason?

doudou: I didn't express myself properly.

The goal, on the inside of autoproj, is to fully get rid of Autoproj.workspace and have a workspace passed explicitely. This is why most of the Autoproj classes (Ops::, CLI::) take a workspace as argument. Without this, we will always be limited to loading one workspace ever in a single process. Which is a bad design for a library.

The one place where Autoproj.workspace is still needed (and will probably always be) are the autobuild files. Outside of the context of autobuild files, the goal is to get rid of Autoproj.workspace.

g-arjones: > Which is a bad design for a library.

Sure, passing `ws` around is a better design for a library. But... is Autoproj a library? I don't see myself regreting I cannot load two workspaces at the same time but I sometimes wish I didn't have to pass `ws` around so often..

That said, making `ws = Autoproj.workspace` does not prevent us from loading two workspaces if we ever need that...

doudou: > I don't see myself regreting I cannot load two workspaces at the same time

I do, actually. There are quite a bit of things that currently can't be done that I wish could be ...

The one problem (and a lot of the work in Autoproj 2.x) was to remove most of the embedding of Autoproj.workspace. The rest of the work is to reduce the reliance on files on disk, instead being more usable.

This is why I make a separation between "to be used in autobuild" and "to be used by autoproj or autoproj tooling"

The former shouldn't have to care about these details.

The latter **should** be passing a workspace explicitly. That's why I don't want to add the default argument here: the caller *should be passing a workspace explicitely*, because he/she should not be relying on Autoproj.workspace.

g-arjones: This is fair enough.. :+1:

PR #278, opened 2019-09-04
fixes for rubocop violations

doudou:


g-arjones:
Appreciate the effort 👍

doudou:
The whole vscode + rubocop setup is a bit fragile when out of a Rock workspace. I just noticed that autoproj-ci has a lot of violations, but that was because the one loaded Autoproj workspace didn't have the vscode gems loaded.

I have to see whether it is possible to have per-folder settings for this stuff.

g-arjones:
I kinda feel like the extension should be setting this up entirely (like we do with c_cpp_properties.json). Not only keeping it in the gem makes the solution fragile as you pointed out, I am not very fond of the idea of having this configuration so far from the only tool that actually uses it...

PR #279, opened 2019-09-05
cli: create 'autoproj test default'

doudou:
Unlike 'enable' or 'disable' without arguments, it only switches
the default without touching the existing exclusions

### DIFF
--- a/lib/autoproj/cli/main_test.rb
+++ b/lib/autoproj/cli/main_test.rb
@@ -22,6 +22,21 @@ module Autoproj
                 end
             end
 
+            desc 'default [on|off]', 'set whether tests are enabled or disabled by default, without touching existing settings'
+            def default(on_or_off)
+                require 'autoproj/cli/test'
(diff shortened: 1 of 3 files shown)


lib/autoproj/cli/main_test.rb
```diff
@@ -22,6 +22,21 @@ def report(report_options = Hash.new)
                 end
             end
 
+            desc 'default [on|off]', 'set whether tests are enabled or disabled by default, without touching existing settings'
+            def default(on_or_off)
```
g-arjones: How about `autoproj enable --default` and `autoproj disable --default` ?

doudou: I thought about it, but liked the explicit `autoproj default`. The reason was that both commands have fundamentally different signatures:

- `autoproj test enable|disable [PACKAGES]`
- `autoproj test default (on|off)`

PR #287, opened 2019-10-03
build reports incrementally

doudou:
Depends on:
- [ ] https://github.com/rock-core/autobuild/pull/78

Apart from being generally nicer (you do get a report even if the build fails in unexpected ways), this ensures that we get a usable report even if the build is terminated hard, for instance
when CI timeouts fail.

In CI context, it avoids having to re-do a complete build and/or test cycle if the CI timeout triggers.


doudou:
Ping @g-arjones

PR #298, opened 2019-12-17
add support to pin an autoproj workspace to a specific bundler version

doudou:
At bootstrap, pass the `--bundler-version` argument.
In an existing workspace, you need to run autoproj_install with
the `--bundler-version` argument.


doudou:
> Looks good, but did not test it myself and a single test on CI fails. But seems to be an issue with the test itself.

Was indeed. Fixed it.

PR #302, opened 2020-02-28
add support for ruby 2.6 and 2.7

doudou:


lib/autoproj/package_set.rb
```diff
@@ -168,7 +168,7 @@ def initialize(
             @version_control = Array.new
             @overrides = Array.new
             @raw_local_dir = raw_local_dir
-            @default_importer = VCSDefinition.from_raw(type: 'none')
+            @default_importer = VCSDefinition.from_raw({ type: 'none' })
```
g-arjones: Is this really necessary? I think rubocop even complains about this

lib/autoproj/vcs_definition.rb
```diff
@@ -307,7 +307,7 @@ def self.call_source_handler(vcs, url, options)
         if !handler
             raise ArgumentError, "there is no source handler for #{vcs}"
         else
-            return handler.call(url, options)
+            return handler.call(url, **options)
```
g-arjones: explicit return statement?

lib/autoproj/package_set.rb
doudou: It's actually one of the major changes. `from_raw`'s signature is `from_raw(specs, <keyword arguments>)`. 2.7 pushes (through a warning) to disambiguate the specs (which is a hash) and the keyword arguments by explicitly adding hash marks.

lib/autoproj/vcs_definition.rb
doudou: I didn't go through the rubocop errors ... just the 2.7 warnings.

doudou:
> Cool.. But tests are failing on 2.7

Yeah ... the point was to have the code reviewed **and** checked on all 4 Ruby versions.

PR #304, opened 2020-03-18
cache: update experimental --gem-compile support

doudou:
Our latest patch for gem-compiler expanded on the initial --artifact
option to allow including and excluding parameters. Update support
in 'cache' to allow for +GLOB and -GLOB specifiers.

The syntax is `GEM_NAME[+INCLUDE][-EXCLUDE][...]`

### DIFF
--- a/lib/autoproj/cli/cache.rb
+++ b/lib/autoproj/cli/cache.rb
@@ -1,9 +1,54 @@
+# frozen_string_literal: true
+
 require 'autoproj/cli/inspection_tool'
 require 'autoproj/ops/cache'
 
 module Autoproj
     module CLI
         class Cache < InspectionTool
+            def parse_gem_compile(string)
+                scanner = StringScanner.new(string)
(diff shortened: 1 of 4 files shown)


lib/autoproj/cli/cache.rb
```diff
@@ -1,9 +1,54 @@
...
+                name = scanner.scan(/[^\[]+/)
+
+                level = 0
+                artifacts = []
+                artifact_include = nil
+                artifact_name = ''.dup
+                until scanner.eos?
+                    c = scanner.getch
+                    if level == 0
+                        raise ArgumentError, "expected '[' but got '#{c}'" unless c == '['
```
g-arjones: How is this possible?

doudou: Syntax error from the user, e.g. `qtbindings[+lib/2.5/*.so.?]-lib/2.5/*.so]`

PR #310, opened 2020-06-04
Add ca-certificates on default.osdeps for archive

### LINKED ISSUE #309: CA certificate update
Hi all!

I have the following source package definition:

```yml
- nanopb:
    retry_count: 3
    type: archive
    url: https://jpa.kapsi.fi/nanopb/download/nanopb-0.3.9.2-linux-x86.tar.gz
```

On my first update try I had the following error:
```bash
OpenSSL::SSL::SSLError (SSL_connect returned=1 errno=0 state=error: certificate verify failed (unable to get local issuer certificate))
```
Which was present because I did not have the updated CA certificates, so I had to install them, in ubuntu `ca-certificates`

My question is, does it make sense to have it added as dependency and always install/update it?

Thanks

ivan1993br:
Adding ca-certificates dependence for archive VC type.

closes #309

### DIFF
--- a/lib/autoproj/default.osdeps
+++ b/lib/autoproj/default.osdeps
@@ -138,15 +138,19 @@ archive:
   debian,ubuntu:
   - tar
   - unzip
+  - ca-certificates
   gentoo:
   - app-arch/tar
   - app-arch/unzip
+  - app-misc/ca-certificates
   arch:
   - tar
   - unzip
+  - ca-certificates
   fedora:
   - tar
   - unzip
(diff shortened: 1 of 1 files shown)


g-arjones:
Did you check the package names on the other distros? Don't add if you are not sure

ivan1993br:
> Did you check the package names on the other distros? Don't add if you are not sure

Yes, but I did not find for macports and homebrew

doudou:
Thanks @ivan1993br

PR #312, opened 2020-07-01
implement autoproj update --ask

doudou:
Depends on:
- [ ] https://github.com/rock-core/autobuild/pull/90

In this mode, all updates have to first be validated by the user
before they are applied.

Autoproj displays the list of changes for each package and asks
whether it should update that package.

It's somewhat slow (because there is less parallelization), but
it improves my update workflow which was so far `status` - `update` 1000x


g-arjones:
Cool

PR #314, opened 2020-07-27
Rescue, specify and reraise empty package name error

0Nel:
A misplaced or spare character (e.g. plus signs or backslashes) are detected by the yaml parser with a useful hint on the respective line number:

    could not find expected '+' while scanning a simple key at line 30 column 3

A misplaces dash on the other hand is only handles in the package_set section. When located within the layout section this causes the manifest.rb to throw a slightly confusing NoMethodError:

      .../manifest.rb:235:in `validate_package_name_argument': undefined method `to_str' for nil:NilClass (NoMethodError)

This PR solely rescues the error and raises a more specific ArgumentError. Alternatively and possibly more general one could check for package.nil? Not sure what's the preferred style here.

      if package.nil?
          raise ArgumentError, "no package name specified. check your manifest file for lines containing only a dash"
      end

### DIFF
--- a/lib/autoproj/manifest.rb
+++ b/lib/autoproj/manifest.rb
@@ -61,6 +61,7 @@ module Autoproj
             end
 
             if data["layout"].member?(nil)
+                Autoproj.warn "There is an empty entry in your layout in #{file}. All empty entries are ignored."
                 data["layout"] = data["layout"].compact
             end
 
--- a/lib/autoproj/manifest.rb
+++ b/lib/autoproj/manifest.rb
@@ -231,11 +231,11 @@ module Autoproj
                 validate_package_in_self(package)
                 package.name
             else
-		begin
+                begin
                     package = package.to_str
-		rescue NoMethodError
-		    raise ArgumentError, "no package name specified. check your manifest file for lines containing only a dash"
-		end
+                rescue NoMethodError
+                    raise ArgumentError, "no package name specified. check your manifest file for lines containing only a dash"
+                end
                 if require_existing && !has_package?(package)
                     raise PackageNotFound, "no package named #{package} in #{self}"
                 end
--- a/lib/autoproj/manifest.rb
+++ b/lib/autoproj/manifest.rb
@@ -60,6 +60,10 @@ module Autoproj
                 YAML.safe_load(File.read(file)) || {}
             end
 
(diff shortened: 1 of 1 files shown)


doudou:
Hey. Could you paste a failing manifest ? It would help understanding how to best handle the error.

g-arjones:
Also, please, don't use tabs :(

0Nel:
@g-arjones sorry, vim autotabbed :(

@doudou a minimal manifest example would be the following:

```
package_sets:
  - github: rock-core/package_set

layout:
  - rock.core
  -
```

It is just about having a proper error message for having an accidental dash anywhere in the layout.

doudou:
I think I'd rather implement it just after loading the manifest file, to make sure it matches the expected format.

0Nel:
Updated the file with regard to @doudou s comment. The loaded manifest is checked for empty entries right away as part of the load(file) function. This is only necessary for the layout section as empty entries in the package-set section already throw an appropriate error. Not sure if it is necessary to warn/inform the user about empty entries?

lib/autoproj/manifest.rb
```diff
@@ -60,6 +60,10 @@ def load(file)
                 YAML.safe_load(File.read(file)) || {}
             end
 
+            if data["layout"].member?(nil)
```
doudou, line 63: You don't need to test for nil. Just call compact.

doudou, line 63: You don't need the test here. Just call compact.

doudou:
> Not sure if it is necessary to warn/inform the user about empty entries?

It would be better, but can be left out of this pull request.

After your comments, I started looking for a way to validate a YAML document loaded from file, which is something that I'm in need of in a lot of places. It turns out that there is JSON Schema that basically allows to validate a hash-of-objects and has a pure ruby implementation. I think it would be a good way forward.

lib/autoproj/manifest.rb
0Nel: Actually added the test in order to print a warning and then decided to go without it :D

Gonna add a brief User warning to the PR.

0Nel:
Sorry for the delay on this one. Finally got the chance to add a small user warning.

doudou (approved):

PR #316, opened 2020-09-13
Add safe navigation to layout read and test for it

### LINKED ISSUE #315: `undefined method `member?'` Error with empty layout
**Environment**
Currently I have a manifest where I concentrate package sets for different projects, it is basically used for CI tests, it has no package on layout.

On this manifest I have no interest to add any packages to be update, is just a package list assembly. If I need one, I specify it.

**Problem**
After this [PR](https://github.com/rock-core/autoproj/pull/314), the empty layout causes a crash, because the `data["layout"]` will not exit and has no `member?` function, as I have no package to update, just the package sets.

**Point of View**
I think that my empty package set is coherent, and should be possible. Although a manifest is mainly used for a build configuration for my workspace, implying that I would need something to build. I do not see a problem to allow empty manifest, for particular cases like this one.

**Solution 1 with current modification with @g-arjones  advice**
I can insert a empty `- ` package on the manifest. I do not like this idea because I think that there is no problem allowing no layouts (which I was doing until now), and the warning makes no sense for my case, I mean, the warning is coherent, but what I want is no package at all.

**Solution 2 another PR**
Check for the existence of the layout before checking the empty entry.

Finally, I will be more then happy to do this new PR if you guys agree with my thoughts. So, what do you think?

ivan1993br:
# Description

This PR basically adds safe navigation for the data["layout"] method call.

Plus, it adds two tests:

- Explicitly check for empty layout load. (I know that the tests were failing because of this call, but I did not see any explicit test for this)
- Warn output if an empty layout is present

I did not went through all the test cases to check if I created a repeated test (there is a lot...). I looked where it made sense for me, but I might be wrong.  

# Issue
closes #315

### DIFF
--- a/lib/autoproj/manifest.rb
+++ b/lib/autoproj/manifest.rb
@@ -60,7 +60,7 @@ module Autoproj
                 YAML.safe_load(File.read(file)) || {}
             end
 
-            if data["layout"].member?(nil)
+            if data["layout"]&.member?(nil)
                 Autoproj.warn "There is an empty entry in your layout in #{file}. All empty entries are ignored."
                 data["layout"] = data["layout"].compact
             end
--- a/test/ops/test_configuration.rb
+++ b/test/ops/test_configuration.rb
@@ -424,6 +424,30 @@ module Autoproj
                 end
             end
 
+            describe "#load_no_packages_layout" do
+                it "load no layout" do
+                    FileUtils.mkdir_p ws.config_dir
+                    manifest_path = File.join(ws.config_dir, 'manifest')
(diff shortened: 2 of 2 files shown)


g-arjones:
Looks good to me. There are some style issues though... Would you please run rubocop (`bundle exec rubocop`) and fix the code you added? You will probably get tons of errors (we are fixing them little by little) just fix the lines you added and that's already good help.

doudou (approved):

doudou:
Sorry @g-arjones  .... didn't see your comment, which is (of course) spot on

PR #322, opened 2021-03-30
generate an env.sh in the prefix and separate build dir roots

doudou:
This allows to spawn a new shell and load the env.sh there.

### DIFF
--- a/lib/autoproj/workspace.rb
+++ b/lib/autoproj/workspace.rb
@@ -807,6 +807,19 @@ module Autoproj
             full_env = self.full_env
             changed = save_cached_env(full_env)
             full_env.export_env_sh(shell_helpers: shell_helpers)
+
+            build_dir_is_absolute = Pathname(build_dir).absolute?
+            full_env.each_env_filename do |_, filename|
+                basename = File.basename(filename)
+                File.open(File.join(prefix_dir, basename), "w") do |io|
+                    io.puts "source \"#{filename}\""
+                end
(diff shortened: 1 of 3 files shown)


lib/autoproj/workspace.rb
```diff
@@ -803,6 +803,19 @@ def export_env_sh(_package_names = nil, shell_helpers: true)
             full_env = self.full_env
             changed = save_cached_env(full_env)
             full_env.export_env_sh(shell_helpers: shell_helpers)
+
+            build_dir_is_absolute = File.absolute_path(build_dir) == build_dir
```
g-arjones, line 807: I think `Pathname#absolute?` is more robust

g-arjones:
I'm wondering what's the use case

doudou:
* open terminal in workspace, autoload env.sh with the aenv script
* go to build dir with acd -b .
* open a new terminal tab, get env.sh autoloaded with the aenv script

lib/autoproj/workspace.rb
doudou: Not sure it's more robust, but it would definitely look better. I'll change.

g-arjones: > Not sure it's more robust

```ruby
File.absolute_path("/root/////") != "/root/////"
```

g-arjones:
> get env.sh autoloaded with the aenv script

Clever :+1:

g-arjones (approved):

PR #324, opened 2021-03-30
test: fix influence of .autoprojrc on Workspace#setup tests

doudou:


g-arjones:
"global" ?

doudou:
Yeah. `load_global_configuration` ?

g-arjones:
That would be better, yes.

PR #325, opened 2021-03-31
Update default python dependencies for debian,ubuntu

2maz:
Currently any pip based package installation on Ubuntu 20.04 fails since autoproj tries to install python-pip - which is not available


ivan1993br:
I was thinking about this, I mean, you are right, python-pip does not exists. However, this dependencies exists because of the `pip autoproj depedency`, that is broken due to the non existence of `pip` on the system, since the `python-...` does not exist anymore, now is `pip3`, and due to this, the `pip  autoproj dependency` module is broken on >20.04. So my question is: should `pip` be updated to `pip3` or should be marked as non existent on newer versions, so the user realize that this required deps needed for autoproj does not exists there? (I mean, for now, following the logic that pip deps is broken on autoproj today).

2maz:
@ivan1993br pip package manager should also be updated. 
I will therefore move functionality from the rock-core package set's python helper into autoproj, mainly to improve the autodetection of the python and corresponding pip version.

2maz:
@ivan1993br I still have to move/add tests, but you might what to give some early feedback. I also added the option to use a python virtualenv.

g-arjones:
Could you please use rubocop to lint your code?

2maz:
Applied rubocop and added tests

g-arjones:
> Applied rubocop 

I still see too many style issue. Did you use [this configuration?](https://github.com/rock-core/autoproj/blob/master/.rubocop.yml). You will probably have to update the ruby target version

2maz:
I think I do but there are some warnings, not sure how critical they are. Note, that I have not been using rubcop before, so I might miss something obvious.

```
roehr:package_managers/ (master✗) $ bundle exec rubocop -l python.rb                                                                                                                             [18:06:14]
warning: parser/current is loading parser/ruby25, which recognizes
warning: 2.5.9-compliant syntax, but you are running 2.5.1.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
/home/dfki.uni-bremen.de/roehr/workspace/rock_master/tools/autoproj/.rubocop.yml: Metrics/LineLength has the wrong namespace - should be Layout
Warning: obsolete parameter `ExcludedMethods` (for `Metrics/BlockLength`) found in /home/dfki.uni-bremen.de/roehr/.autoproj/gems/ruby/2.5.0/gems/rubocop-rock-0.6/defaults.yml
`ExcludedMethods` has been renamed to `IgnoredMethods`.
Warning: obsolete parameter `ExcludedMethods` (for `Metrics/BlockLength`) found in /home/dfki.uni-bremen.de/roehr/workspace/rock_master/tools/autoproj/.rubocop.yml
`ExcludedMethods` has been renamed to `IgnoredMethods`.
The following cops were added to RuboCop, but are not configured. Please set Enabled to either `true` or `false` in your `.rubocop.yml` file.

Please also note that can also opt-in to new cops by default by adding this to your config:
  AllCops:
    NewCops: enable
...
```

```
bundle exec rubocop --version                                                                                                                                [18:09:30]
warning: parser/current is loading parser/ruby25, which recognizes
warning: 2.5.9-compliant syntax, but you are running 2.5.1.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
1.12.1
```

g-arjones:
Don't use the `-l` option

2maz:
There should only be one issue remaining:
```
test_python.rb:105:20: C: [Correctable] Style/HashEachMethods: Use each_value instead of values.each.
                op.values.each do |p|

```

but with the suggested fix I would otherwise get

```
  1) Error:
activate_python#test_0005_does update the python path:
NoMethodError: undefined method `each_value' for #<Autobuild::Package::EnvOp:0x000056526bc24c28>
Did you mean?  each_slice
    package_managers/test_python.rb:105:in `block (2 levels) in <module:Python>'
```

2maz:
any further feedback?

ivan1993br:
Sorry to take so long to give a feedback here. It is a feature that was really in need. I have just some small observations:

- The Python class is not a `PackageManager` and all others in the folder are. Shouldn't it be in other location? I think it is a useful class that have some functions which may be used on some other modules like some implementation of `SetupTools` on `autobuild`
- When I add a pip dependency with `USE_PYTHON=False` `autoproj` breaks raising an error.
<details>
  <summary>Details</summary>

```
Traceback (most recent call last):
	18: from /home/ivan/dev/src/tools/autoproj/bin/aup:17:in `<main>'
	17: from /home/ivan/.local/share/autoproj/gems/ruby/2.7.0/gems/thor-1.1.0/lib/thor/base.rb:485:in `start'
	16: from /home/ivan/.local/share/autoproj/gems/ruby/2.7.0/gems/thor-1.1.0/lib/thor.rb:392:in `dispatch'
	15: from /home/ivan/.local/share/autoproj/gems/ruby/2.7.0/gems/thor-1.1.0/lib/thor/invocation.rb:127:in `invoke_command'
	14: from /home/ivan/.local/share/autoproj/gems/ruby/2.7.0/gems/thor-1.1.0/lib/thor/command.rb:27:in `run'
	13: from /home/ivan/dev/src/tools/autoproj/lib/autoproj/cli/main.rb:240:in `update'
	12: from /home/ivan/dev/src/tools/autoproj/lib/autoproj/cli/main.rb:128:in `run_autoproj_cli'
	11: from /home/ivan/dev/src/tools/autoproj/lib/autoproj/reporter.rb:107:in `report'
	10: from /home/ivan/dev/src/tools/autobuild/lib/autobuild/reporting.rb:103:in `report'
	 9: from /home/ivan/dev/src/tools/autoproj/lib/autoproj/reporter.rb:110:in `block in report'
	 8: from /home/ivan/dev/src/tools/autoproj/lib/autoproj/cli/main.rb:139:in `block in run_autoproj_cli'
	 7: from /home/ivan/dev/src/tools/autoproj/lib/autoproj/cli/update.rb:138:in `run'
	 6: from /home/ivan/dev/src/tools/autoproj/lib/autoproj/workspace.rb:861:in `install_os_packages'
	 5: from /home/ivan/dev/src/tools/autoproj/lib/autoproj/os_package_installer.rb:434:in `install'
	 4: from /home/ivan/dev/src/tools/autoproj/lib/autoproj/os_package_installer.rb:434:in `each'
	 3: from /home/ivan/dev/src/tools/autoproj/lib/autoproj/os_package_installer.rb:435:in `block in install'
	 2: from /home/ivan/dev/src/tools/autoproj/lib/autoproj/os_package_installer.rb:445:in `install_manager_packages'
	 1: from /home/ivan/dev/src/tools/autoproj/lib/autoproj/package_managers/pip_manager.rb:48:in `install'
/home/ivan/dev/src/tools/autoproj/lib/autoproj/package_managers/pip_manager.rb:32:in `guess_pip_program': uninitialized constant Autoproj::PackageManagers::PipManager::ValueError (NameError)
```
</details>

When set to `True` it goes smoothly

Anyways, anxious to the merge!

2maz:
@ivan1993br  thanks for the feedback

> The Python class is not a PackageManager and all others in the folder are. Shouldn't it be in other location? I think it is a useful class that have some functions which may be used on some other modules like some implementation of SetupTools on autobuild

I moved it to autoproj/python.rb as first step. I assume you also want me to split the functionality and move version detection and all parts not depending on Workspace and Config to autobuild. I would prefer to wait and see if that is really needed.

> When I add a pip dependency with USE_PYTHON=False autoproj breaks raising an error.

That was a wrong exception with no test coverage. Added a test case to pip_manager.

2maz:
ping

g-arjones:
Sorry for taking so long @2maz. I will try to have a look until next week

lib/autoproj/package_managers/pip_manager.rb
```diff
@@ -25,39 +27,53 @@ def os_dependencies
...
+                        raise ConfigError, "Your current package selection" \
+                          " requires the use of pip is required, but" \
+                          " the use of python has been denied, see" \
+                          " setting of USE_PYTHON in your workspace configuration." \
+                          " Either remove all packages depending on pip packages " \
+                          " from the workspace layout (manifest) or " \
+                          " call 'autoproj reconfigure' to change the setting."
+                    end
+                else
+                    Autoproj::Python.setup_python_configuration_options(ws: ws)
```
g-arjones, line 41: You should error out here if the user chose not to use python...

lib/autoproj/python.rb
```diff
@@ -0,0 +1,317 @@
...
+
+          cmd = "#{python_bin} -m virtualenv #{venv_name}"
+          msg, status = Open3.capture2e(cmd)
+          unless status.success?
+              raise "Autoproj::Python.prepare_venv preparation of"\
+                  " virtual env '#{venv_name}' failed: '#{msg}'"
+          end
+
+          ws.env.set('VIRTUAL_ENV_DISABLE_PROMPT', 1)
+          ws.env.source_after File.join(Autoproj.root_dir, venv_name, "bin", "activate")
```
g-arjones, line 282: This, unfortunately, won't integrate well. `Environment#source_after` will just add a source line to the end of the generated `env.sh` file but that file should only be treated as a convenience for the user and not as a dependency. In other words, running `.autoproj/bin/autoproj` should work as expected even if you don't pre-load `env.sh`. This is important for external tools that rely on an autoproj workspace.

If we want to integrate a "virtual env" tool (which I think is a good idea) I think you should look into [pipenv](https://pypi.org/project/pipenv/) which works more like Ruby's Bundler but then it would probably be best to implement a new `PackageManagers::PipenvManager` and deprecate/remove the current `PackageManagers::PipManager`.

Also, since we are already installing pip packages locally in the workspace, what do we get with the virtualenv integration that you implemented?

lib/autoproj/python.rb
```diff
@@ -0,0 +1,317 @@
...
+          end
+
+          cmd = "#{python_bin} -m virtualenv #{venv_name}"
+          msg, status = Open3.capture2e(cmd)
+          unless status.success?
+              raise "Autoproj::Python.prepare_venv preparation of"\
+                  " virtual env '#{venv_name}' failed: '#{msg}'"
+          end
+
+          ws.env.set('VIRTUAL_ENV_DISABLE_PROMPT', 1)
```
g-arjones, line 281: What is this for? I don't see it being used anywhere

lib/autoproj/python.rb
```diff
@@ -0,0 +1,317 @@
...
+          File.join(Autoproj.root_dir, venv_name)
+        end
+
+        def self.setup_python_configuration_options(ws: Autoproj.workspace)
+            ws.config.declare 'USE_PYTHON', 'boolean',
+                              default: 'no',
+                              doc: ["Do you want to activate python?"]
+
+            if ws.config.get("USE_PYTHON")
+                ws.os_package_installer.install(%w[python pip])
```
g-arjones, line 292: This is not necessary. `PipManager` [has an internal dependency on `pip` already](https://github.com/rock-core/autoproj/blob/master/lib/autoproj/package_managers/pip_manager.rb#L24)

g-arjones (changes requested):

g-arjones:
I didn't know why (didn't have time to debug) but the configuration is not being saved.

Also, the use of the shims is inconsistent when virtual env is enabled:

```
arjones@781567b66194:~/test$ which python
/home/arjones/test/.autoproj_python_venv/bin/python
arjones@781567b66194:~/test$ autoproj which python
/home/arjones/test/install/bin/python
arjones@781567b66194:~/test$ 
```

Which is yet another reason to not use virtualenv at all, IMHO.

lib/autoproj/package_managers/pip_manager.rb
2maz: What exactly you mean? Replace "raising ConfigError"? Currently, a ConfigError is raised when a preference exists at all, i.e. USE_PYTHON is set?

g-arjones: What I mean is that this should not happen:

```
arjones@781567b66194:~/test$ autoproj osdeps
  Do you want to activate python? [no] no
  installing/updating Python dependencies: mbed_cli, mbed_tools
  updated environment
Command finished successfully at 2021-05-30 18:04:30 -0300 (took 5 secs)
arjones@781567b66194:~/test$ 
```

lib/autoproj/python.rb
2maz: You do not see the special prompt from virtualenv, since it disables the prompt.

lib/autoproj/python.rb
2maz: While I have overseen these given dependencies in PipManager, I wonder under what circumstances these dependencies are installed?

lib/autoproj/python.rb
g-arjones: Then shouldn't this always be set? The current implementation sets this only when `USE_PYTHON` does not exist

lib/autoproj/python.rb
g-arjones: They get installed when there's a package in the workspace that would be installed by the given package manager. All `PackageManagers` subclasses implement this method.

2maz:
> Also, the use of the shims is inconsistent when virtual env is enabled:

Why is this a problem?

> Which is yet another reason to not use virtualenv at all, IMHO.

How would pipenv handle this?

lib/autoproj/python.rb
2maz: Sorry, but I cannot follow you on this one? It is only set, when the virtual env is in use. At least that would be the intention of the current code, and this should only be the case when USE_PYTHON is set?

lib/autoproj/python.rb
2maz: Thanks, for pointing this out. I am not entirely sure, however, if this is a condition that is fulfilled for all users?

lib/autoproj/python.rb
g-arjones: The current code sets `VIRTUAL_ENV_DISABLE_PROMPT` [when `setup_python_configuration_options` is called](https://github.com/rock-core/autoproj/blob/40fcf063f8c62f3b8e29ae3b48de0cd5510d47e8/lib/autoproj/package_managers/pip_manager.rb#L41).  But that only gets called if `USE_PYTHON` does not exist (it won't be called if `USE_PYTHON` exists AND is `true`)

lib/autoproj/python.rb
g-arjones: Not sure I get your point. Pip should only be installed if the user needs pip (aka wants to install a package with pip), right?

g-arjones:
> Why is this a problem?

Don't you expect that, within the same workspace, the same python binary is used regardless of whether a python script is being run internally by autoproj, by doing `autoproj exec my_script.py` or  by just running the script directly?

> Which is yet another reason to not use virtualenv at all, IMHO.

As far as I know, you don't have to source an env script for pipenv to work. `pipenv python` will give you the proper python binary. I don't have a lot of experience with `pipenv` either so you should probably look into the documentation to figure out why people are moving away from virtualenv.

lib/autoproj/python.rb
2maz: > Also, since we are already installing pip packages locally in the workspace, what do we get with the virtualenv integration that you implemented?

I assumed this is a more 'python'-conforming setup and thus improving the integration, but I do not consider me as the python expert here. I will have a look at pipenv.

g-arjones: Yeah, but the current code does not integrate well... So, if there's no reason to use virtualenv I think we should either use something that integrates better or just don't virtualenv.

2maz:
> Don't you expect that, within the same workspace, the same python binary is used regardless of whether a python script is being run internally by autoproj, by doing `autoproj exec my_script.py` or by just running the script directly?

Sure. But I do not follow your argument, when you are just calling 'python ...' and can assume that the PATH setup is consistent?
Isn't this more related to the problem that you refer to with ws.env.source_after

g-arjones:
> Isn't this more related to the problem that you refer to with ws.env.source_after

Probably, yes. But even you fix that somehow, you will still have two shims pointing to different python binaries which may break current setups (i.e for users that have the shim from `prefix/install/bin` as the shebang operator in shell scripts or are somehow expecting that shim to be valid).

2maz:
@g-arjones sorry it took me so long to come back to this: I removed now the venv part and updated according to the remaining suggestions.

lib/autoproj/package_managers/pip_manager.rb
```diff
@@ -25,39 +27,49 @@ def os_dependencies
...
 
+            # rubocop:disable Lint/UnusedMethodArgument
             def install(pips, filter_uptodate_packages: false, install_only: false)
                 guess_pip_program
-                if pips.is_a?(String)
-                    pips = [pips]
-                end
+                pips = [pips] if pips.is_a?(String)
 
-                base_cmdline = [Autobuild.tool('pip'), 'install','--user']
```
g-arjones, line 41: Isn't `--user` required to install in `PYTHONUSERBASE` ?

g-arjones:
Tests are failing...

2maz:
Sorry for bothering you with the previous version. Updated -- tests and rubocop should pass now.

g-arjones:
@2maz Sorry, but `test_python.rb` is still failing.

@doudou Master is broken as well, btw. Did you run autoproj's test suite when you implemented rock-core/autobuild#101? I did not investigate but at first glance, it seems like either that or maybe rock-core/autobuild#96 itself broke `Autoproj::Ops::Cache` and bootstraping-related tests.

2maz:
> Sorry, but test_python.rb is still failing.

Mhm, not for me. Where/how does it fail?

g-arjones:
> Mhm, not for me. Where/how does it fail?

```
arjones@arjones-eos:~/dev/autoproj$ git rev-parse HEAD
5a339917f0c18915b04b81335bbb3f3b85ecc926
arjones@arjones-eos:~/dev/autoproj$ bundle exec ruby -Itest test/test_python.rb 
Run options: --seed 50061

# Running:

.E.  WARN: Autoproj::Python.rewrite_python_shims: creating /tmp/d20210615-6045-6ptm0w/install/bin - are you operating on a valid autoproj workspace?
....  WARN: Autoproj::Python.rewrite_python_shims: creating /tmp/d20210615-6045-3c9mv0/install/bin - are you operating on a valid autoproj workspace?
.

Finished in 1.710720s, 4.6764 runs/s, 26.8893 assertions/s.

  1) Error:
activate_python#test_0006_does not update python path:
Autoproj::UserError: not in a Autoproj installation
    /home/arjones/dev/autoproj/lib/autoproj/system.rb:41:in `root_dir'
    /home/arjones/dev/autoproj/lib/autoproj/workspace.rb:923:in `workspace'
    /home/arjones/dev/autoproj/lib/autoproj/python.rb:250:in `activate_python_path'
    test/test_python.rb:133:in `block (2 levels) in <module:Python>'

8 runs, 46 assertions, 0 failures, 1 errors, 0 skips
arjones@arjones-eos:~/dev/autoproj$ 
```

2maz:
@g-arjones thanks. Apparently there was a mix up with an available workspace. Just fixed.

g-arjones (approved):

g-arjones:
@2maz Thanks for the effort, I think this is OK to merge.

I would like to make one last comment though.

The more I look at this patch the more I wonder whether we are doing more than we should. I mean, all the heuristics/guesswork to determine python version, activation code and writing shims are there just to support python2 which has officially reached its end-of-life. Personally, if I were the one fixing the original issue (wrong dependencies on Ubuntu 20.04+) I would have just fixed the osdeps, defaulted `Autobuild.programs["pip"] = "pip3"`, `Autobuild.programs["python"] = "python3"` and told users to use that API if they needed/wanted python2 for whatever reason. Just food for thought.

Unless @doudou has anything to say, feel free to merge.

doudou:
> Unless @doudou has anything to say, feel free to merge.

No comments on my side, Apart from that, no, I haven't run autoproj tests with the latest autobuild changes..

I honestly know too little about Python dev tools to be able to contribute meaningfully.

doudou (approved):

2maz:
> The more I look at this patch the more I wonder whether we are doing more than we should. I mean, all the heuristics/guesswork to determine python version, activation code and writing shims are there just to support python2 which has officially reached its end-of-life. A significant amount of code. Personally, if I were the one fixing the original issue (wrong dependencies on Ubuntu 20.04+) I would have just fixed the osdeps, defaulted Autobuild.programs["pip"] = "pip3", Autobuild.programs["python"] = "python3" and told users to use that API if they needed/wanted python2 for whatever reason. Just food for thought.

Actually tried to improve the pip manager here with code that was residing in the rock package set. IMO the change provides a self-explanatory and transparent mechanism for users to pick the python version. I also prefer using the configuration over direct API usage since this gives a streamlined and better tractable workflow for the setup.

lib/autoproj/python.rb
```diff
@@ -0,0 +1,290 @@
...
+                    python_bin, = auto_resolve_python(ws: ws)
+                end
+
+                ws.config.declare 'python_executable', 'string',
+                                  default: python_bin.to_s,
+                                  doc: ["Select the path to the python executable"]
+
+                activate_python(ws: ws)
+
+                ws.config.declare 'USE_PYTHON_VENV', 'boolean',
```
g-arjones, line 281: I think you forgot to remove this

g-arjones:
> I also prefer using the configuration

How does the user pick the version during the proposed configuration mechanism?

2maz:
> How does the user pick the version during the proposed configuration mechanism?

by picking the corresponding python binary

lib/autoproj/python.rb
2maz: Unfortunately yes, thx. Updated.

g-arjones:
Feel free to merge... Thanks.

doudou:
@2maz could you write a paragraph about the new Python support ? What it does and some basic documentation on how to use it. It would be for the release notes for the next Autoproj version.

2maz:
How about the following:

Multiple Python versions can now coexist in a system, although not in a single autoproj workspace.
For this purpose autoproj users can now be explicitly asked for any needed Python support during a bootstrap, by adding the following to the init.rb:
```
    require 'autoproj/python'
    Autoproj::Python.setup_python_configuration_options()
```
In case Python support is requested or implicitly needed, e.g., following a pip dependency in a layout, the user is asked to provide the path to the Python executable to use. By default the Python executable is searched/guessed automatically.
Corresponding shims are installed for an active workspace: install/bin/python and install/bin/pip.

To ensure a consistent setup for the generation of language bindings, package definitions in *.autobuild file can be extended as follows:
```                
cmake_package 'custom/pkg_with_python_bindings' do |pkg|
    bin, version, sitelib_path = Autoproj::Python.activate_python_path(pkg)
    pkg.define 'PYTHON_EXECUTABLE', bin if bin
    pkg.define 'PYTHON_VERSION', version if version
end
```
The above setting will only be used when the user has requested Python support, or in general, when USE_PYTHON is set to true in the autoproj configuration (see .autoproj/config.yml).

doudou:
Great ! Thanks

PR #333, opened 2021-09-08
feat: allow installing gems from git

doudou:
Built on top of #330


doudou:
This is the last bit needed to have a proper 20.04 bootstrap. See https://github.com/rock-core/package_set/pull/194

I checked that it works as advertised on both 18.04 and 20.04. Now, the osdeps parsing code does little validation, so it might be that it would break some cases (hopefully low likelihood given the amount of unit tests around it, though)

g-arjones (approved):

PR #335, opened 2021-10-11
Fix package set sorting

### LINKED ISSUE #328: sanity check in sort_package_sets_by_import_order fails in some cases
The following exception is triggered in some cases:
https://github.com/rock-core/autoproj/blob/700ae181206fbb6bb3560bf914e423582d8c8a11/lib/autoproj/ops/configuration.rb#L370

My theory is, root_pkg_set / "main configuration" is inserted just after the last explicit import, but there can be further imported dependencies if the last explicit package set has dependencies. Somehow the implicit dependencies seem to not being propagated properly to the root package set. A workaround is to add root pkg set also to the "target list" 

https://github.com/vbargsten/autoproj/commit/363008fab3c675f36664f62f4d3be7fe048fe8ca#diff-86d73a3e07d681adc4f4e2efba488f6709905a27818933f9a89b02cf2cbafbb2

2maz:
Fixes #328 and corrects the load order of init.rb files to respect the import order of package sets

fyi @moooeeeep

### DIFF
--- a/lib/autoproj/ops/configuration.rb
+++ b/lib/autoproj/ops/configuration.rb
@@ -1,57 +1,51 @@
+require "rgl/adjacency"
+require "rgl/dot"
+require "rgl/topsort"
+
 module Autoproj
     module Ops
         #--
         # NOTE: indentation is wrong to let git track the history properly
         #+++
 
-        # Comparator can be used to compare the list of package sets
-        # so that they can be sorted according to the following constraints:
-        # 1. dependencies among the package sets
-        # 2. order in which they are defined in the main configuration (manifest)
-        #
-        # Note that 1. naturally always takes precedence over 2. since 2.
-        # only affects a sorting of package sets that have no dependency
-        class Comparator
-            attr_reader :imports_of, :natural_import_order
+        # PackageSetHierachy be used to build the hierarchy of package set
+        # imports, as directed acyclic graph (DAG)
+        # so that they can be (topologically) sorted according to their
+        # dependencies
+        class PackageSetHierarchy
+            attr_reader :dag
 
             def initialize(package_sets, root_pkg_set)
-                @imports_of = Hash.new
+                @dag = RGL::DirectedAdjacencyGraph.new
+
                 package_sets.each do |p|
-                    @imports_of[p] = PackageSet.resolve_imports(p)
+                    p.imports.each do |dep|
+                        @dag.add_edge dep, p
+                    end
                 end
 
-                @natural_import_order = Array.new
-                root_pkg_set.imports.each do |p|
-                    @imports_of[p] = PackageSet.resolve_imports(p)
-                    @imports_of[p].each do |import|
-                        unless @natural_import_order.include?(import)
-                            @natural_import_order << import
-                        end
+                @dag.add_vertex root_pkg_set
+                import_order = root_pkg_set.imports.to_a
+                import_order.each_with_index do |p, index|
+                    if index + 1 < import_order.size
+                        @dag.add_edge p, import_order[index + 1]
+                        @dag.add_edge p, root_pkg_set
                     end
-                    @natural_import_order << p unless @natural_import_order.include?(p)
+                end
+
+                unless @dag.acyclic?
+                    raise "The package set hierarchy contains cycles: #{@dag.cycles}"
                 end
             end
 
-            def validate_import_order(pkg_set)
-                imports = pkg_set.imports.to_a
-                until imports.empty?
-                    pkg_set = imports.shift
-                    imports.each do |p|
-                        if @imports_of[pkg_set].include?(p)
-                            raise "PackageSet '#{pkg_set.name}' depends on '#{p.name}', "\
-                                  "but is listed before it in the manifest"
-                        end
-                    end
-                end
+            # Flatten the hierarchy, a establish a sorting
+            def flatten
+                @dag.topsort_iterator.to_a
             end
 
-            def compare(pkgset_a, pkgset_b)
-                return 0 if pkgset_a == pkgset_b
-                return 1 if imports_of[pkgset_a].include?(pkgset_b)
-                return -1 if imports_of[pkgset_b].include?(pkgset_a)
-
-                # Check manifest precedence for pkgset_a != pkgset_b
-                @natural_import_order.index(pkgset_a) <=> @natural_import_order.index(pkgset_b)
+            # Write the hierarchy to an image (png) file
+            def to_png(path)
+                @dag.write_to_graphic_file("png", path.gsub(".png", ""))
             end
         end
 
@@ -368,18 +362,13 @@ module Autoproj
             # Package sets that have no dependencies come first,
             # the local package set (by main configuration) last
             def sort_package_sets_by_import_order(package_sets, root_pkg_set)
-                c = Comparator.new(package_sets, root_pkg_set)
-                unordered_pkg_sets = package_sets + [root_pkg_set]
-                sorted_pkg_sets = unordered_pkg_sets.uniq.sort do |a, b|
-                    c.compare(a, b)
-                end
+                c = PackageSetHierarchy.new(package_sets, root_pkg_set)
+                sorted_pkg_sets = c.flatten
 
                 if sorted_pkg_sets.last != root_pkg_set
                     raise InternalError, "Failed to sort the package sets: the " \
                                          "root package set should be last, but is not #{sorted_pkg_sets.map(&:name)}"
                 end
-
-                c.validate_import_order(root_pkg_set)
 
                 sorted_pkg_sets
             end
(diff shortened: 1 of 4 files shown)


lib/autoproj/ops/configuration.rb
```diff
@@ -157,10 +157,10 @@ def create_remote_set_user_dir(vcs)
                 symlink_dest
             end
 
-            def load_package_set(vcs, options, imported_from)
+            def load_package_set(vcs, options, imported_from, initrb: true)
```
doudou, line 160: It seems that `load_package_set` is never used outside of `load_and_update_package_sets`. Now that you load init.rb within `update_package_sets`, the loading here should be removed.

lib/autoproj/ops/configuration.rb
```diff
@@ -314,56 +314,56 @@ def inspect
...
-                    end
-                    if queue.size == last_size
-                        raise ArgumentError, "cannot resolve the dependencies between package sets. There seem to be a cycle amongst #{queue.map(&:name).sort.join(', ')}"
-                    end
+                name_to_pkg_set = Hash.new
+                pkg_sets = Hash.new
+                all_dependencies = []
+
+                package_sets.each do |p|
+                    name_to_pkg_set[p.name] = p
```
doudou, line 326: Is going through names really needed ? I would think that you should be using the package set object directly here (which would remove the need for `name_to_pkg_set`).

lib/autoproj/ops/configuration.rb
```diff
@@ -314,56 +314,56 @@ def inspect
...
-                    dependencies = pkg_set.imports.dup
-                    if dependencies.empty?
-                        result.unshift(pkg_set)
-                    else
-                        i = result.find_index do |p|
-                            dependencies.delete(p)
-                            dependencies.empty?
+                unknown_pkg_sets = all_dependencies - pkg_sets.keys
+                unless unknown_pkg_sets.empty?
+                    raise "Unknown package sets found: #{unknown_pkg_sets}"
```
doudou, line 337: InternalError ?

lib/autoproj/ops/configuration.rb
```diff
@@ -314,56 +314,56 @@ def inspect
...
+                end
+
+                sorted_pkg_sets = []
+                pending_pkg_sets = pkg_sets.keys
+
+                until pending_pkg_sets.empty?
+                    add = []
+                    progress = false
```
doudou, line 345: I believe that progress == true is the same as !add.empty? (i.e. no need for an extra flag)

lib/autoproj/ops/configuration.rb
```diff
@@ -314,56 +314,56 @@ def inspect
...
+                    pending_pkg_sets.each do |pkg_set_name|
```
doudou, line 346: This is

~~~ ruby
add = pending_pkg_sets.find_all do |pkg_set_name|
   pkg_sets[pkg_set_name] -= sorted_pkg_sets
   pkg_sets[pkg_set_name].empty?
end
~~~

lib/autoproj/ops/configuration.rb
```diff
@@ -314,57 +313,50 @@ def inspect
...
-                    end
-                    if queue.size == last_size
-                        raise ArgumentError, "cannot resolve the dependencies between package sets. There seem to be a cycle amongst #{queue.map(&:name).sort.join(', ')}"
-                    end
+                name_to_pkg_set = Hash.new
+                pkg_sets = Hash.new
+                all_pkgset_deps = Set.new
+
+                package_sets.each do |p|
+                    pkg_sets[p.name] = p.imports.to_a
```
doudou, line 325: My initial remark was related to `pkg_sets` too ... You could (I believe "should") use `p` as the key here.

doudou: Also, to avoid two loops, maybe `(package_sets.to_a + [root_pkg_set]).each` (arguable)

lib/autoproj/package_set.rb
```diff
@@ -792,6 +832,24 @@ def resolve_overrides(key, vcs)
             vcs
         end
 
+        # Recursively resolve imports for a given package set
+        def self.resolve_imports(pkg_set, parents = Set.new)
+            return Set.new if pkg_set.imports.empty?
+
+            updated_parents = parents.dup
+            updated_parents << pkg_set
```
doudou, line 840: updated_parents = parents | [pkg_set]

lib/autoproj/package_set.rb
```diff
@@ -4,6 +4,45 @@ module Autoproj
...
     # dependencies that are provided by the operating system (.osdeps file).
     class PackageSet
+        # Comparator can be used to compare the list of package sets
+        # so that they can be sorted according to the following constraints:
+        # 1. dependencies among the package sets
+        # 2. order in which they are defined in the main configuration (manifest)
+        #
+        # Note that 1. naturally always takes precedence over 2. since 2.
+        # only affects a sorting of package sets that have no dependency
+        class Comparator
```
doudou, line 14: This is really not an algorithm/property of the package sets, but of the overall configuration. Leave it in ops/configuration.rb

lib/autoproj/package_set.rb
```diff
@@ -4,6 +4,45 @@ module Autoproj
...
+            def compare(pkgset_a, pkgset_b)
+                return 0 if pkgset_a == pkgset_b
+                return 1 if pkg_sets[pkgset_a].include?(pkgset_b)
+                return -1 if pkg_sets[pkgset_b].include?(pkgset_a)
+
+                # Check manifest precedence for pkgset_a != pkgset_b
+                return 0 unless main_imports.include?(pkgset_a)
+                return 0 unless main_imports.include?(pkgset_b)
+
+                if main_imports.index(pkgset_a) > main_imports.index(pkgset_b)
```
doudou, line 38: main_imports.index(pkgset_a) <=> main_imports.index(pkgset_b)

lib/autoproj/package_set.rb
```diff
@@ -4,6 +4,45 @@ module Autoproj
...
     class PackageSet
+        # Comparator can be used to compare the list of package sets
+        # so that they can be sorted according to the following constraints:
+        # 1. dependencies among the package sets
+        # 2. order in which they are defined in the main configuration (manifest)
+        #
+        # Note that 1. naturally always takes precedence over 2. since 2.
+        # only affects a sorting of package sets that have no dependency
+        class Comparator
+            attr_reader :pkg_sets, :main_imports
```
doudou, line 15: Would you consider renaming `pkg_sets` into something like `imports_of` ? The name has really confused me.

lib/autoproj/package_set.rb
doudou: resolve_imports does make sense here though

lib/autoproj/ops/configuration.rb
```diff
@@ -314,57 +313,20 @@ def inspect
...
-                        result.unshift(pkg_set)
-                    else
-                        i = result.find_index do |p|
-                            dependencies.delete(p)
-                            dependencies.empty?
-                        end
-                        result.insert(i + 1, pkg_set)
-                    end
+                c = PackageSet::Comparator.new(package_sets, root_pkg_set)
+                sorted_pkg_sets = c.pkg_sets.keys.sort do |a, b|
```
doudou, line 321: Better use `package_sets + [root_pkg_set]` here. Using `c.pkg_sets.keys` here feels like an anti-pattern to me. Or define a `all_package_sets` on Comparator and use that.

doudou (approved):

lib/autoproj/ops/configuration.rb
```diff
@@ -1,9 +1,70 @@
...
+        # dependencies
+        class PackageSetHierarchy
+            attr_reader :imports_of, :dag
+
+            def initialize(package_sets, root_pkg_set)
+                @dag = RGL::DirectedAdjacencyGraph.new
+
+                @imports_of = Hash.new
+                package_sets.each do |p|
+                    dependencies = PackageSet.resolve_imports(p)
```
doudou, line 23: We don't need the recursive resolve_imports anymore, since the DAG does it naturally. Just iterate over all package sets and add edges for each direct dependency.

lib/autoproj/ops/configuration.rb
```diff
@@ -1,9 +1,70 @@
...
+                    pkg_set = imports.shift
+                    imports.each do |p|
+                        if @imports_of[pkg_set].include?(p)
+                            raise "PackageSet '#{pkg_set.name}' depends on '#{p.name}', "\
+                                  "but is listed before it in the manifest"
+                        end
+                    end
+                end
+            end
+
```
doudou, line 56: The acyclic? test already takes care of what this was doing ... Maybe remove ? (which would get rid of @imports_of)

lib/autoproj/ops/configuration.rb
2maz: It does only so if all the involved package sets are provided (or the tree depth is limited) and that does not seem to be guaranteed at that point?

doudou: It is guaranteed at this point. I've created a commit to document the whole class, but I think I'll propose it as a new PR after this one gets merged.

doudou:
@2maz ... All good for me, feel free to merge

PR #336, opened 2021-10-11
Distribution specific override

2maz:
This PR outlines an approach to add distribution specific overrides in each package set, by linking into the main configuration overrides.d folder.
If the taken approach is fine with you @doudou @g-arjones, then I will add tests.


2maz:
> The overrides that get added by this mechanism will be applied at the very end of everything. I.e. a package set that's "downstream" to e.g. rock-core would be overriden by stuff from rock-core. I think it will make for a very fragile mechanism.

If overrides are applied in import order, I do not see a problem?!

> I would personally have preferred to add a osdeps-like structure within source.yml, which would at least be consistent with the osdeps system, and would allow to apply the same patterns than within the osdeps (e.g. have rules that apply to both debian & ubuntu, group rules for different versions and then have a default ...)

Thought about it, but I still think that having overrides in separate file/folder structure makes the handling easier

lib/autoproj/package_set.rb
```diff
@@ -510,9 +511,52 @@ def load_description_file
...
+            overrides = []
             if (data = source_definition["overrides"])
-                [[source_file, data]]
+                overrides << [source_file, data]
+            end
+
+            pkgset_overrides_dir = File.join(raw_local_dir, "overrides.d")
+            return overrides unless File.exist?(pkgset_overrides_dir)
+
+            overrides += load_overrides_from_dir(pkgset_overrides_dir)
```
doudou, line 524: Shouldn't this be sorted by name somewhere ?

PR #341, opened 2021-11-11
feat: add support for direnv

doudou:
This patch auto-generates a .envrc file along with env.sh, which
simply sources env.sh. This allows to use direnv, which adds env
unloading when we quit a folder (nice !)


doudou:
Actually, direnv does not support function definitions - it loads the env file in a subshell to see what it loads.

So, not working for autoproj

PR #344, opened 2021-11-15
Preserve environment when running scripts that require sudo in ShellS…

g-arjones:
…criptManager

### DIFF
--- a/lib/autoproj/package_managers/shell_script_manager.rb
+++ b/lib/autoproj/package_managers/shell_script_manager.rb
@@ -3,7 +3,8 @@ module Autoproj
         # Base class for all package managers that simply require the call of a
         # shell script to install packages (e.g. yum, apt, ...)
         class ShellScriptManager < Manager
-            def self.execute(command_line, with_locking, with_root, env: Autobuild.env)
+            def self.execute(command_line, with_locking, with_root,
+                env: Autobuild.env, inherit: Set.new)
                 if with_locking
                     File.open("/tmp/autoproj_osdeps_lock", "w") do |lock_io|
                         until lock_io.flock(File::LOCK_EX | File::LOCK_NB)
@@ -25,8 +26,9 @@ module Autoproj
                     process_env.add_path("PATH", "/usr/local/sbin",
                                          "/usr/sbin", "/sbin")
 
+                    inherit.each { |var| process_env.set(var, env[var]) }
                     sudo = Autobuild.tool_in_path("sudo", env: process_env)
-                    command_line = [sudo, *command_line]
+                    command_line = [sudo, "--preserve-env", *command_line]
                 end
 
                 Autobuild::Subprocess.run "autoproj", "osdeps", *command_line,
@@ -201,7 +203,8 @@ module Autoproj
             # @return [Boolean] true if packages got installed, false otherwise
             def install(packages, filter_uptodate_packages: false, install_only: false,
                 auto_install_cmd: self.auto_install_cmd,
-                user_install_cmd: self.user_install_cmd)
+                user_install_cmd: self.user_install_cmd,
+                inherit: Set.new)
                 return if packages.empty?
 
                 handled_os = ws.supported_operating_system?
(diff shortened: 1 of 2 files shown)


doudou:
When not used with `isolate_environment`, this is rather dangerous.

Given that the "users" of this class are internal stuff, I think it would be best that the caller provides what needs to be forwarded explicitly.

g-arjones:
> I think it would be best that the caller provides what needs to be forwarded explicitly.

Could you elaborate? My use case is installing an apt dpkg package that requires setting an env var to accept an user agreement/license.

Having the user messing with the AptDpkgManager or OSPackageManager for that sounds too involved just to set an env var

g-arjones:
I've changed to make it go through `AptDpkgManager` so at least we know how this is being used internally.

But to be honest,  I don't think it's safer. Actually, the vulnerability I think you are concerned about (environment injection) is already there and this patch does not make it worse. If an attacker wants to inject something, all it takes is writing a `sudo` wrapper and appending `--preserve-env` to the command line since we are already using `Autobuild.env` to find `sudo` anyway.

doudou:
> since we are already using Autobuild.env to find sudo anyway.

Good point. We should only look for it in /bin and /usr/bin by default.

I'm still feeling not-so-warm feelings about this. We already have specific flags for this in BundlerManager (to pass build arguments to gems), something equivalent for AptDpkgManager, that lists the environment variables that can be passed ? The package set that defines the problematic osdep can add it in its init.rb.

g-arjones:
I think this is as safe as we can reasonably get. It's a breaking change though since every manager gets an isolated environment and every script that runs without root gets `Autobuild.env` (previously got `ENV`, which was probably a bug)

lib/autoproj/package_managers/shell_script_manager.rb
```diff
@@ -3,7 +3,8 @@ module PackageManagers
         # Base class for all package managers that simply require the call of a
         # shell script to install packages (e.g. yum, apt, ...)
         class ShellScriptManager < Manager
-            def self.execute(command_line, with_locking, with_root, env: Autobuild.env)
+            def self.execute(command_line, with_locking, with_root,
+                env: Autobuild.env, inherit: Set.new)
```
doudou, line 7: use `ws.env` instead of `Autobuild.env`

g-arjones: ?

As the default keyword argument?

doudou:
> I think this is as safe as we can reasonably get. It's a breaking change though since every manager gets an isolated environment and every script that runs without root gets `Autobuild.env` (previously got `ENV`, which was probably a bug)

The ENV->ws.env is breaking in principle, but I agree that it was probably a bug.

The other change modifies behavior for every manager subclassing ShellScriptManager that runs as root. And, for those, will actually add environment variables that before were not defined *if the caller requires it*. Should not change current behavior, unless I am mistaken.

g-arjones:
> Should not change current behavior

I don't think so either.

lib/autoproj/package_managers/shell_script_manager.rb
doudou: Yes

g-arjones: Then you mean `Autoproj.workspace.env`, right?

doudou: Nope. Manager has a `#ws` attribute. I've been trying to avoid these globals when doing autoproj 2.0

g-arjones: > Manager has a #ws attribute

Yes, but `self.execute` is a class method and `#ws` is an instance method.

doudou: Ah ... got it. Sorry. Autobuild.env was already there actually ... :p I thought you added it.

Autoproj.workspace.env would probably be better, but the really good one would have been to be mandatory. Oh well.

doudou (approved):

PR #346, opened 2021-11-16
declare python as a regular dependency on python packages

g-arjones:
An internal dependency triggers installation unnecessarily during build.


g-arjones:
Actually, python is required during import, this is not a proper fix.

PR #347, opened 2021-11-18
setup python after importing python packages

g-arjones:
Depends on:
  - [x] https://github.com/rock-core/autobuild/pull/112


g-arjones:
@doudou Any idea why python shims cannot be found during load when running without sourcing the environment file (using autoproj shims from dot_autoproj_dir)? Apparently `Package#update_environment` is called before autoproj's `PATH` is setup.

g-arjones:
Nevermind... Found the issue.

doudou (approved):

2maz (approved):

PR #348, opened 2021-11-18
filter osdeps up-to-date packages during build

### LINKED ISSUE #340: Autoproj keeps trying to install python3-dev when a python_package is selected

g-arjones:
Fixes #340

### DIFF
--- a/lib/autoproj/ops/import.rb
+++ b/lib/autoproj/ops/import.rb
@@ -467,7 +467,9 @@ module Autoproj
 
                 raise failures.first if !keep_going && !failures.empty?
 
-                install_internal_dependencies_for(*all_processed_packages)
+                install_internal_dependencies_for(
+                    *all_processed_packages, install_only: import_options[:checkout_only]
+                )
                 finalize_package_load(all_processed_packages, auto_exclude: auto_exclude)
 
                 all_enabled_osdeps = selection.each_osdep_package_name.to_set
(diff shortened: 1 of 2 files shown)


doudou:
It feels like a weird side-effect. Build already sets `checkout_only` to true, which definitely should not try to reinstall a package that's already there.

I rather think there's something wrong with the apt-dpkg detection of the package.

g-arjones:
Not sure I get your point... `checkout_only` is not handled by `AptDpkgManager`. Autoproj always try to install internal dependencies during build, it's up to the manager to guess whether the dependency needs to be installed or not and `AptDpkgManager` only does that if either `filter_uptodate` or `install_only` are `true`.

doudou:
`checkout_only` on `CLI::Update` should be translated into `install_only` for the osdeps

- https://github.com/rock-core/autoproj/blob/master/lib/autoproj/cli/update.rb#L105
- definition of `normalize_osdep_options`: https://github.com/rock-core/autoproj/blob/master/lib/autoproj/cli/update.rb#L185

I think that the problem is with the `install_only` logic of AptDpkgManager, which does not detect the package is already installed while `filter_uptodate_package` follows a different logic and does.

g-arjones:
> checkout_only on CLI::Update should be translated into install_only for the osdeps

Yeah, that would make sense there's a bug somewhere in there. I've just printed `install_only` in `AptDpkgManager` and that is `false` there. I'll dig some more.

g-arjones:
Found it... Good hint.

doudou (approved):

PR #351, opened 2021-11-20
enable tests in python packages

g-arjones:
Depends on:
  - [ ] https://github.com/rock-core/autobuild/pull/113


doudou:
Does it really make sense to choose a test framework globally like this ? The Python community does not have multiple test harnesses ?

g-arjones:
Either `pytest` and `unittest` but `pytest` can be used just as a runner for `unittest` tests (and other frameworks as well). So, AFAIK, `pytest` should fit most use cases.

g-arjones:
Turns out there are limitations in using pytest as a runner for other frameworks. Since I don't have enough python experience to be sure whether these are relevant or not, I'm closing this PR. I will be glad to add it back if more experienced python people are interested.

PR #353, opened 2021-11-23
Add python package extension to activate python

g-arjones:
Follow-up from: https://github.com/rock-core/autoproj/pull/352

This also fixes a bug in the python activation code that was extracted from rock's package set where shims wouldn't be written if the user tried to set `USE_PYTHON` in his local configuration.


doudou (approved):

g-arjones:
I've been using this for a while now and since python support is essentially broken without this, I'm merging.

PR #355, opened 2021-11-25
ShellScriptManager: make sure to not inherit original env

g-arjones:
Turns out that `Autobuild::Subprocess#run` still inherits the original env even when an environment is passed as an argument. Pass `env_inherit: false` to fix that. Also add tests that really should have been added in the first PR that touched this.


doudou (approved):

doudou:
> Also add tests that really should have been added in the first PR that touched this.

We all need a regular reminder that tests are actually useful, every once in a while ;-)

PR #356, opened 2021-12-03
fix: allow upgrades in ruby minor version

doudou:


lib/autoproj/package_managers/bundler_manager.rb
```diff
@@ -532,23 +535,26 @@ def install(gems, filter_uptodate_packages: false, install_only: false)
                 # Save the osdeps entries in a temporary gemfile and finally
                 # merge the whole lot of it
                 gemfile_contents = Tempfile.open "autoproj-gemfile" do |io|
+                    io.puts "source \"https://rubygems.org\""
```
g-arjones, line 538: Shouldn't you be checking whether  `gems ` is empty before doing this? I wonder if `Bundler::Dsl.evaluate` will give you yet another warning telling that you have a Gemfile with a source but no gems.

doudou: Not sure it's necessary ... but seems to me that it is a good idea regardless. Will do.

g-arjones:
@doudou Ping?

g-arjones (approved):

PR #361, opened 2021-12-17
Extend ros manifests

g-arjones:


lib/autoproj/condition_parser.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
+
+require "parslet"
+require "autoproj/exceptions"
+require "autoproj/variable_expansion"
+
+module Autoproj
+    # Parses a conditional expression
+    # Syntax and rules as defined in https://www.ros.org/reps/rep-0149.html#id20
+    class ConditionParser < Parslet::Parser
```
doudou, line 10: Since syntax is derived from a ROS REP, I'd prefer that you call this ROSConditionParser

g-arjones: Done!

doudou (approved):

PR #369, opened 2022-02-17
fix: not complete package_sets list during execution of init.rb

annaborn:
This is a fix for an issue, that we encountered after updating to autoproj 2.15.1.

In 2.15.1, the package set's init.rb was called when the manifest's package list was still not complete. 

In our case, package set's init.rb relies on the following information:

```
...
root_pkg_set = ws.manifest.main_package_set
...
```

Since the main package set was loaded as the last, "ws.manifest.main_package_set" is nil::Class at this point.

To resolve this issue, run each package's init.rb file after all package sets have been loaded into the manifest.

### DIFF
--- a/lib/autoproj/ops/configuration.rb
+++ b/lib/autoproj/ops/configuration.rb
@@ -626,12 +626,18 @@ module Autoproj
                 root_pkg_set.imports.each do |pkg_set|
                     pkg_set.explicit = true
                 end
+
+                # sort packages, main package is the last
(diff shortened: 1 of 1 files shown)


annaborn:
@doudou Hello Sylvain! Could you please check the fix?

doudou (approved):

doudou:
@annaborn sorry for the delay. Released as 2.15.2.

Issue #364, opened 2022-02-07
Feature discussion: Possibility to auto-apply a seed_config file from the buildconf repo


planthaber:
For our CI Pipeline it would be beneficial to store a seed_config.yml in our buildconf repository and to auto-apply it when calling autoproj_bootstrap when the file is present in the buildconf repo.

Currently we have to download the seed_config file first and then call autoproj_bootstrap with the parameter.

I think the best option would be to add a paramater `--seed-config-from-buildconf=PATH`

@doudou What do you think about this idea, or is there already a way to do this? 

I case you think this is possible, we'll try to provide PR for this featue.

FYI: @0Nel @skasperski

g-arjones:
I think the seed configuration is used during install (before checking out the buildconf). Why not use the `init.rb` ?

planthaber:
Yes, you can call autoproj_bootstrap with --seed-config=PATH, but is has to be locally available before calling the script.

I'd like to use it from the "to be checked out" buildconf.

Using the init.rb might be possible, but this will the the default for all bootstraps and not only if set.

One could add an configuration option to the init.rb to ask if defaults should be used though.

Is there a call to load the config seed from the init.rb?

This would solve our issue, as we can then provide a common config_seed via command line for all boostraps on our build server, that just  sets the "load default config_seed option" to true when calling autoproj_bootstrap

planthaber:
> > Yes, you can call autoproj_bootstrap with --seed-config=PATH, but is has to be locally available before calling the script.
> 
> What I meant was that what you are suggesting doesn't sound feasible (at least not without a lot of changes) because autoproj needs the seed configuration before checking the buildconf out so `--seed-config-from-buildconf` doesn't make much sense.
> 
> > One could add an configuration option to the init.rb to ask if defaults should be used though.
> 
> Either that or you can make it dependent on an env var that you would set in your CI or even a constant in the main manifest (then you would have one manifest for the CI and one for development).

But this seed would also be applied after cloning the buidlconf.

As far as i understoot autoproj_bootstrap it installs autoproj and dependencies and then calls 
the bootstrap.rb of the installed autoproj and passes the params here: https://github.com/rock-core/autoproj/blob/master/bin/autoproj_bootstrap#L834

I thought if would only be adding the parameter and then merging the config from the buildconf repo after the buildconf was checked out and before aup is run.
[...]

PR #372, opened 2022-03-08
Configuration: Add option to automatically load a config seed from a buildconf

planthaber:
Allows automatic loading a default "seed-config.yml" file from the buildconf repository
```ruby
Autoproj.config.load_config_once('seed-config.yml')
```
in a buildconf init.rb to apply a configuration seed from the same buildconf repository

The second call will ask the user if the default should be applied or not:
```ruby
Autoproj.config.load_config_once_with_permission('seed-config.yml')
```
This will load the autoproj/seed-config.yml while bootstrapping (to avoid questions asked)

Discussed here: https://github.com/rock-core/autoproj/issues/364


doudou (commented):
Since this is an instance method, it should rather obviously act on itself instead of Autoproj.config. And `configuration_option` as well as `user_config` are now instance methods on the same class (named declare and get I think)

Instead of referring to the global Autoproj.workspace, let the caller provide you with a `root_path` keyword argument whose default is `Autoproj.workspace.root`

lib/autoproj/configuration.rb
```diff
@@ -606,5 +606,34 @@ def to_hash
...
+        # Allows to load a seed-config.yml file (also from a buildconf repository)
+        # rather than providing it before checkout using the --seed-config paramater
+        # of the autoproj_bootstrap script
+        # this allows to bootstrap with --no-interactive and still apply a custom config e.g. in CI/CD
+        # The call to this function has to be in the init.rb of the buildconf BEFORE any other
+        # config option, e.g. the git server configuration settings
+        # The filename parameter is the name of the config seed yml file in the repository
+        def load_config_once(filename)
+            unless Autoproj.config.has_value_for?("default_config_applied_#{filename}")
+                seed_config = File.join(Autoproj.workspace.root_dir, "autoproj", filename)
```
doudou, line 619: Do not create massive assumptions on where `filename` is. Using `File.expand_path(filename, root_dir)` would allow the caller to provide a full path. Where root_dir's default is `Autoproj.workspace.config_dir`

lib/autoproj/configuration.rb
```diff
@@ -606,5 +606,34 @@ def to_hash
...
+
+        # Allows to load a seed-config.yml file (also from a buildconf repository)
+        # rather than providing it before checkout using the --seed-config paramater
+        # of the autoproj_bootstrap script
+        # this allows to bootstrap with --no-interactive and still apply a custom config e.g. in CI/CD
+        # The call to this function has to be in the init.rb of the buildconf BEFORE any other
+        # config option, e.g. the git server configuration settings
+        # The filename parameter is the name of the config seed yml file in the repository
+        def load_config_once(filename)
+            unless Autoproj.config.has_value_for?("default_config_applied_#{filename}")
```
doudou, line 618: Do an early return instead of a gigantic if.

~~~
return if get("default_config_applied_#{filename}", false)

<rest of method>
~~~

doudou:
Please write unit tests for this ...

planthaber:
Updated the PR (added tests and added the suggested stuff)

I realized that due to the "load once" rule the user is asked if the config should be applied when calling "autoproj reconfigure" but it is not applied because the config said is was already applied.

I think this is actually "good" behavior, otherwise the default could overwrite question asked before the defaults are applied, so I now prevent asking for that option when calling reconfigure

lib/autoproj/configuration.rb
```diff
@@ -606,5 +606,35 @@ def to_hash
...
+        # rather than providing it before checkout using the --seed-config paramater
+        # of the autoproj_bootstrap script
+        # this allows to bootstrap with --no-interactive and still apply a custom config e.g. in CI/CD
+        # The call to this function has to be in the init.rb of the buildconf BEFORE any other
+        # config option, e.g. the git server configuration settings
+        # The filename parameter is the name of the config seed yml file in the repository
+        def load_config_once(filename)
+            return if get("default_config_applied_#{filename}", false)
+
+            seed_config = File.expand_path(filename, Autoproj.workspace.config_dir)
```
doudou, line 620: What I meant in my earlier review is that the root_dir should be passed as a keyword argument, that is

```
def load_config_once(filename, root_dir: Autoproj.workspace.config_dir)
```

test/test_configuration.rb
```diff
@@ -433,5 +433,61 @@ module Autoproj
                 assert loaded_config.get(option_b_name) == b_value
             end
         end
+
+        describe "#load_config_once" do
+            it "load config once loads config only once" do
```
doudou, line 438: Please add a test for the root_dir argument (i.e. resolution of relative paths)

doudou (approved):

planthaber:
Updated the PR

doudou:
In this case, I would expect to be the one merging as I and @g-arjones are the de-facto maintainer of this package. If you had a lot of contributions, I would expect you to merge.

doudou:
All good. Thanks !

PR #376, opened 2022-07-13
Improve load_config_once_with_permission documentation

planthaber:


doudou:
hi @planthaber. CI does not pass because Rubocop changed a config name. Would you mind renaming the entry to make it pass ? That would be very helpful.

planthaber:
Now it seems to work:

Deleted the "Gemspec/DateAssignment" cop 

and

Fixed "lib/autoproj/os_package_installer.rb:188" issue by calling "nil" (noop), based on https://stackoverflow.com/questions/13452624/equivalent-of-pass-in-ruby

PR #379, opened 2022-08-02
Don't depend on ignored packages

g-arjones:
Ignored packages are not imported and removed from selection during build (its Rake tasks are not created). However, they're still added as dependencies of other packages resulting in autobuild invoking Rake tasks that do not exist and a failure that a regular user is unlikely to make sense of.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ group :vscode do
     gem "debase", ">= 0.2.2.beta10"
     gem "pry"
     gem "pry-byebug"
-    gem "rubocop", ">= 0.6.0"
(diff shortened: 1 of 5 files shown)


doudou:
It would be great to add a system test, that is running update and build on a workspace with an ignored package, to make sure this does not get broken again.

doudou (approved):

g-arjones:
@doudou Not sure this is what you meant... What do you think?

doudou:
I think that's already a lot better than the current situation :+1:

doudou (approved):

Issue #381, opened 2022-09-06
Support conditional dependencies in manifests


pierrewillenbrockdfki:
Conditional dependencies would help solve the problem where a package can be built in multiple ways but depending on different sets of packages for each of those ways.

The actual problem here are packages that can be configured to compile either with qt4 or with qt5. Currently, a manifest would have to include both with `<depend package="qt4" optional="1" /> <depend package="qt5" optional="1" />`, which makes autoproj try to install both qt4 and qt5, if available, and it will try to build the package even when the correct qt(as configured) could not be installed.

Having autoproj install both qt4 and qt5 for a workspace that only actually requires qt4 is not desirable, and attempting a build that is destined to fail due to missing dependency is not desirable either.

In https://github.com/rock-core/package_set/pull/210#issuecomment-1199870798, @doudou proposed new manifest syntax:
`<depend ... if="..." />`

This would probably be a good solution, so lets try to flesh it out a bit more.

## Compatibility

Is compatibility with rosdep needed, that will parse the manifest.xml if given the opportunity, or has that already been dropped?

Older autoproj would interpret the `<depend ... if="..." />` syntax like a normal always-on dependency, so is there merit to add a variant that older autoproj would always ignore?: `<conddepend ... if="..." />`

## The Condition

The condition would have to be something that can be set by package-sets. Maybe support expressions.

I would try to have a ruby expression here with terms that are predefined from a per-package hash. These hashes could then be manipulated by package-sets.

Before i start implement something based on eval here, is there a better option?

g-arjones:
You can actually do that already if you use ROS manifests (`package.xml`) instead. 

[See the manifest syntax](https://github.com/rock-core/autoproj/blob/eb572a36b78ec36aba31cf6efc534534677be8e6/test/test_ros_package_manifest.rb#L175)
[See the expression syntax](https://github.com/rock-core/autoproj/blob/master/test/test_ros_condition_parser.rb)

The implementation follows [ROS' REP-149](https://ros.org/reps/rep-0149.html#dependency-tags).

I think that's something reasonable to have in autoproj's vanilla manifest as well and if you decide to implement it, please do it in a way that you reuse as much code as possible.

pierrewillenbrockdfki:
> You can actually do that already if you use ROS manifests (`package.xml`) instead.
> 
[...]

PR #382, opened 2022-09-06
Add "condition" attribute to "depend" tag of autoproj manifest.xml

pierrewillenbrockdfki:
As discussed in #381 , all the infrastructure for this is already in place. The only missing piece is actually evaluating the new attribute in the context of autoproj manifest.xml.


lib/autoproj/package_manifest.rb
```diff
@@ -185,6 +185,14 @@ def initialize(path, manifest)
                 super()
                 @path = path
                 @manifest = manifest
+                @env = manifest.package.ws.env
+                @condition_parser = RosConditionParser.new(@env)
```
doudou, line 189: I'm having a conceptual issue with using the environment for this. It would force us to expose the variables in the environment.

I would very much prefer that we limit ourselves to the configuration variables.

test/test_package_manifest.rb
```diff
@@ -155,6 +156,14 @@ def parse_dependency(xml)
                     dependency = parse_dependency("<package><depend package='test' optional='1'/></package>")
                     assert dependency.optional
                 end
+                it "is not included if the condition is false" do
+                    manifest = PackageManifest.parse(pkg, "<package><depend package='test' condition='1==0'/></package>")
```
doudou, line 160: You are only testing constants, not that you can evaluate variables that have been set in the chosen variable source (i.e. env or conf)

pierrewillenbrockdfki: More complex expressions are already tested in test_ros_condition_parser.rb.

lib/autoproj/package_manifest.rb
pierrewillenbrockdfki: This is exactly the same as what is done for the ros package.xml. But i am open for suggestions for what context to use instead of manifest.package.ws.env. I will try to find where the configuration variables are stored.

test/test_package_manifest.rb
pierrewillenbrockdfki: > More complex expressions are already tested in test_ros_condition_parser.rb.

But yeah, should also test if a variable actually manages to work its way from (in this case) manifest.package.ws.env into the condition parser.

doudou: That was my point exactly ;-)

doudou:
> The Configuration is further filtered to remove anything that is not a String

Not sure I understand the reason ?

pierrewillenbrockdfki:
> > The Configuration is further filtered to remove anything that is not a String
> 
> Not sure I understand the reason ?

As far as i can tell, the ros_condition_parser only deals in strings, since that is the only thing they will see in the manifest.package.ws.env. Seeing that arrays are present in the config-seed.yml, i assume that those are also present in manifest.package.ws.config, so filtering those out to document that anything but strings does not work seems to be a good idea. If this turns out to be problematic with numbers or booleans, something needs to be done with those in any case.

pierrewillenbrockdfki:
Converted this to a draft, i want to see if i can avoid the dance for getting at the workspace, either by making the context a parameter to be passed in our by moving the dependency filtering to a later stage.

doudou:
> If this turns out to be problematic with numbers or booleans, something needs to be done with those in any case.

I think we definitely should allow to test for booleans. Numbers could be just converted to string, but booleans really can't (I would expect to be able to test for `condition=someboolean`

pierrewillenbrockdfki:
> > If this turns out to be problematic with numbers or booleans, something needs to be done with those in any case.
> 
> I think we definitely should allow to test for booleans. Numbers could be just converted to string, but booleans really can't (I would expect to be able to test for `condition=someboolean`

I do agree. I also think that using string comparison on numbers produces unintuitive results("10" < "2").

For the ROS package.xml, the current implementation is exactly as required by the [REP-149](https://ros.org/reps/rep-0149.html#dependency-tags)("All variables are substituted by their values and treated as strings." and "the comparison operators only do a string comparison and don't attempt to interpret the string as a numerical value").

So, options here are to partially duplicate the ros_condition_parser.rb, enhance it so that the comparison operators can be provided externally, or adopt the REP-149 rules for the autoproj manifest.xml conditions. This should better be decided now, since it is different externaly visible behavior.

pierrewillenbrockdfki:
The latest commit changes the way package_manifest.rb and ros_package_manifest.rb get at their respective condition_context. I moved that up to the Manifest class, because it always has a usable workspace available, where the package_manifests need to go through the package passed to them to get at the workspace, or use the Autoproj.workspace.

This leaves the package_manifests with less hidden dependencies, as can be seen by the changes that could be done to the tests. I did not add tests for if the environment or configuration find their way to the package_manifests, yet. At a later point, i want to squash these commits again to clean them up.

@doudou do you think this is the variant to continue with, or do you like the previous way more?

doudou:
I like the change. I think we could convert numbers and booleans to strings. Checking a boolean will still need to use == "true" and == "false", but at least it is feasible. That would keep the REP rule's spirit I think.

The cherry on top of the cake would be to allow accessing values in configuration hashes by converting them to a string with `.` separators, e.g. `{ "some" => { "config" => 42 } }` would be available as $some.config == "42". I would understand if you feel you've done enough already though ;-)

pierrewillenbrockdfki:
This additionally pulls the variable expansion function out of the RosConditionParser. This is so the values can be run through the Configuration#get method, which can ask the user for a value if none has been given, yet. Just using Configuration#to_hash does not allow this and would have required to discern the Configuration#declare values from those that have been Configuration#set (which can be done also). Additionally, the operation_system value is special cased to make it useful in a condition with only stringly compare available.

lib/autoproj/package_manifest.rb
```diff
@@ -181,10 +184,48 @@ class Loader < BaseLoader
...
+                    _, os_versions = os
+                    return "" unless os_versions.any? do |ver|
+                        var == "operating_system_version_#{ver.gsub(/[.,+-]/, '_')}"
+                    end
+                    return "true"
+                end
+
+                config.get(var)
+            rescue Autoproj::ConfigError
+                ""
```
doudou, line 213: I'm not sure about that one. I would expect the build to fail if a condition is not available. It would avoid people starting to randomly wonder why things don't build ...

pierrewillenbrockdfki: I can allow the exception to bubble up, the error message seems to be informative enough in that case: `ERROR: undeclared option 'BAR'`. I will repackage the error message in an Autobuild::PackageException, like is done for REXML::ParseException(and fix the use of the uninitialised "file" variable).

doudou:
Neat. Just waiting to check for the CI and I merge it. Thanks a lot.

pierrewillenbrockdfki:
> Neat. Just waiting to check for the CI and I merge it. Thanks a lot.

Not sure if there is anything that can be done for
```
 test/test_ros_package_manifest.rb:143:13: C: Metrics/BlockLength: Block has too many lines. [27/25]
            Autoproj::RosPackageManifest::Loader::SUPPORTED_MODES.each do |mode| ...
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
that one got its now overly high count of lines due to lines getting too long and then having to split each keyworded argument to its own line.

lib/autoproj/ros_package_manifest.rb
```diff
@@ -23,10 +23,13 @@ class Loader < PackageManifest::Loader
...
                              exec_depend test_depend run_depend doc_depend].to_set.freeze
 
-            def initialize(path, manifest)
+            def initialize(path, manifest, condition_context: {})
                 super
-                @env = manifest.package.ws.env
-                @condition_parser = RosConditionParser.new(@env)
+                @condition_parser = RosConditionParser.new do |var|
+                    Autoproj.expand(var, condition_context)
+                rescue StandardError
```
g-arjones, line 30: Since you removed this from the condition parser could you please add a test case to `test_ros_package_manifest` that ensures unset variables evaluate to empty strings in ROS manifests as expected by ROS packages? You've added the expander to the condition parser's test suite and you are testing it there at the same time which is kinda pointless.

g-arjones (changes requested):

lib/autoproj/ros_package_manifest.rb
pierrewillenbrockdfki: > You've added the expander to the condition parser's test suite and you are testing it there at the same time which is kinda pointless.

You are right, moved that specific test up to test_ros_package_manifest, testing that behavior through the manifest functions.

g-arjones: Thanks

g-arjones:
There are rubocop failures.

pierrewillenbrockdfki:
> There are rubocop failures.

Fixed the one about the extra space in {}, which brings us back to my original question about what to do with the block around a number of testcases that now has too many lines.

doudou:
> Fixed the one about the extra space in {}, which brings us back to my original question about what to do with the block around a number of testcases that now has too many lines.

In this particular case, I think it would be fair to disable the check for the block. Just add `# rubocop:disable Metrics/BlockLength` at the end of the line that fails.

pierrewillenbrockdfki:
> Just add `# rubocop:disable Metrics/BlockLength` at the end of the line that fails.

Also added Metrics/LineLength, the disabling comment pushes that over the edge.

doudou (approved):

doudou:
@g-arjones ?

g-arjones (approved):

g-arjones:
Looks good to me.

pierrewillenbrockdfki:
I was on vacation, then had other work to do; i see this is still not merged, this time, the tests failed due to an error from bundler regarding debase. Can someone of you @doudou @g-arjones take a look and maybe rerun?

doudou:
Hey @pierrewillenbrockdfki. Sorry to not have pushed forward. There is indeed an unrelated issue that blocks CI. I did fix the underlying problem (https://github.com/rock-core/autoproj/pull/386), but then there seem to have an heisenbug / new bug that I did not have time to try and solve.

doudou:
@pierrewillenbrockdfki it should be fixed now ... but I'm not sure whether you will have to rebase to get the fix. And I can't find a way to manually re-trigger the build. Maybe you have to do it since you are the pull request owner ?

planthaber:
As author of https://github.com/rock-core/autoproj/pull/385 i can tell that the author can re-run the test manually, but it doesn't help with the fix, i just merges the current master to the branch: That helped and re-ran the tests automatically (because of new push).

pierrewillenbrockdfki:
> As author of #385 i can tell that the author can re-run the test manually, but it doesn't help with the fix, i just merges the current master to the branch: That helped and re-ran the tests automatically (because of new push).

Yes, that merges the fix for the github CI system("fix: bundle install without 'vscode'").
As far as this PR, someone else with permissions("maintainer") needs to trigger CI, because of being a "first-time contributor".

pierrewillenbrockdfki:
So, turns out disabling Metric/LineLength seems to not be needed for CI, only for my local rubocop(granted, that one complains alot about what i assume configuration it cannot read).

Thank you to whoever pushed the button, but you'll have to push it again.

pierrewillenbrockdfki:
CI seems to be happy, @doudou , want to merge this?

doudou:
Sure, thanks for the patience !

PR #384, opened 2022-10-24
fix: only requires an 'id' for keyservers definitions

caioaamaral:
For "url" definitions, we don't need an `id` key, e.g:

```yml
- type: key
  url: "https://dl.cloudsmith.io/public/caddy/stable/gpg.key"
- type: repo
  repo: "deb https://dl.cloudsmith.io/public/caddy/stable/deb/debian any-version main"

```

### DIFF
--- a/lib/autoproj/repository_managers/apt.rb
+++ b/lib/autoproj/repository_managers/apt.rb
@@ -285,19 +285,21 @@ module Autoproj
             end
 
             def validate_key_definition(definition)
-                if definition["id"].nil?
-                    raise ConfigError, "#{INVALID_REPO_MESSAGE}: 'id' key missing"
-                elsif !definition["id"].is_a?(String)
-                    raise ConfigError, "#{INVALID_REPO_MESSAGE}: 'id' should be a String"
-                elsif definition["url"] && definition["keyserver"]
-                    raise ConfigError,
-                          "#{INVALID_REPO_MESSAGE}: 'url' conflicts with 'keyserver'"
+                if definition.has_key?("keyserver") # only keyserver uses 'id'
+                    if definition["id"].nil?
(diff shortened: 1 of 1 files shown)


g-arjones:
If you remove the `id` autoproj will keep downloading the keyfile every time you update. The id is required so autoproj can check whether the key was already added.

caioaamaral:
> 

can we use the "url" itself for this check?

g-arjones:
Currently, we check if the key with the given id already exists in the keychain, so using a randomly generated fingerprint doesn't make sense (since it will obviously not be in the keychain).

If you really mind having to add the key fingerprint to the `.repos` file I guess you could cache the keyfile somewhere in the workspace and compare its md5 hash with the remote file but that means you will still have to hit the remote server every time (which may be undesirable) and you'll have to handle `update --only-local`. I really don't think the effort is worth it but you're welcome to try.

@doudou Thoughts?

caioaamaral:
> Currently, we check if the key with the given id already exists in the keychain, so using a randomly generated fingerprint doesn't make sense (since it will obviously not be in the keychain).

What if I just set the `id`, internally in autoproj, with the same value stored at the `url`? I believe it would still be a unique id, compatible with the keychain checking.

I'm ok with manually setting a key directly on the yml, but wouldn't that be fragile somehow? For example, a user could set the same id for two different keys (unless there is an internal check for that already)

g-arjones:
> What if I just set the id, internally in autoproj, with the same value stored at the URL?

I don't understand what you are suggesting. What do you mean by the value stored at the URL?

> For example, a user could set the same id for two different keys

I don't think we should worry about that because there's nothing we can do. The user can still do that when getting the key from a keyserver. If two entries have the same fingerprint then the keys are the same and the current behavior is appropriate in my opinion (only add one of them).

PR #385, opened 2022-10-27
Add more verbose error output if osdep is undefined for an os

planthaber:
Now also prints the location on the definition for other the os, making it easier to add the definition for the new os at the same place


doudou:
Sorry ... this got lost.

PR #387, opened 2022-12-22
fix for configuration_option ArgumentError with ruby3

vbargsten:
Unsure if this is the intended way, but it works for me on Ubuntu 22.04. Problem is described here: https://rubyreferences.github.io/rubychanges/3.0.html#keyword-arguments-are-now-fully-separated-from-positional-arguments, i.e. number of arguments wrong, 3 given expected 2.


doudou (approved):

doudou (commented):
You are hitting one of these difficult backward compatible changes in some cases but not in others ...

You are calling Autoproj.configuration_option passing a hash as third argument, which is why you only need the ** in config.declare. Other people that call configuration_option with keyword syntax will need the ** there as well, but that might break your use case.

Could you try it out ? Adding ** in front of options line 33 ?

My apologies for the very late reply. Hectic year.

vbargsten:
That looks indeed more consistent. I don't see a difference. However, it could break callers using a hash as argument (and not **hash transforming it into keyword arguments).

pierrewillenbrockdfki:
@doudou does this look right? CI seems to be failing due to bundler now hard requiring ruby 3. This seems to work for `autoproj status` on ubuntu 22.04/ruby 3, ubuntu 20.04/ruby 2.7, ubuntu 18.04/ruby 2.5

chhtz:
Wouldn't replacing `Autoproj.configuration_option` by `Autoproj.config.declare` everywhere also resolve this issue? (I made some PRs for that)

PR #389, opened 2022-12-27
fix: auto-select bundler 2.3.6 on Ruby 2.5

doudou:


g-arjones:
Is Ruby 2.5 really the only version affected by the bundler bug?

doudou:
This commit does not fix the bundler bug. It fixes the permanent bug, which is that 2.4.x is not compatible with ruby 2.5

g-arjones (approved):

Issue #401, opened 2023-12-19
Autoproj tries to install incompatible Bundler version


chhtz:
As of version 2.5, bundler requires Ruby version 3.0, which is not (easily) available on some systems.
Therefore bootstraps will fail.
You can work-around this by adding the following line to your config_seed.yml (if you use that):

    bundler_version: '2.4.22'

But maybe there is a more elegant solution (also not fixing the bundler version for users who use Ruby 3 already).

doudou:
Another temporary soluition is to put

```
Autoproj.config.set "bundler_version", "2.4.22"
```

in autoproj/init.rb, which ensures that everyone that use your buildconf gets the fix.

moooeeeep:
Does it make sense to add some logic [to the determination of the default bundler version](https://github.com/rock-core/autoproj/blob/master/bin/autoproj_bootstrap#L762), to account for the new version constraints?

Untested example:

    def self.default_bundler_version
        ruby_version = Gem::Version.new(RUBY_VERSION)
        if ruby_version < Gem::Version.new("2.7.0")
            # e.g., Ubuntu 18.04
            return "2.3.6"
        elsif ruby_version < Gem::Version.new("3.0.0")
            # e.g., Ubuntu 20.04
            return "2.4.22"
        else
            # e.g., Ubuntu 22.04
            return
        end
    end

PR #402, opened 2024-01-04
fix: pin bundler version on ruby < 3

g-arjones:
Fixes https://github.com/rock-core/autoproj/issues/401


moooeeeep:
Thanks! With this I can bootstrap on Ubuntu 20.04 again (Ubuntu 22.04 seems to work too).

doudou (approved):

Issue #400, opened 2023-12-01
make .git suffix in repo url optional


annaborn:
Hello  Sylvain @doudou.

Can we make a .git suffix in repo url optional? The repositories in azure devops does not follow the standards, so their repo url does not have .git at the end. We have now a trouble to checkout the azure repos.
I mean this line
https://github.com/rock-core/autoproj/blob/47e93f159d7053c99e8b6f93139da11b56e695cb/lib/autoproj/git_server_configuration.rb#L103

Does .git suffix required anyway? Since the checkout/push/pull in gitlab and github seems to work also without .git.

We can provide a pull request, just need to know what solution would be accepted. :)

Thank you in advance!

doudou:
It's a bit meaningless, the best would obviously to just tell everyone to add `.git` to their URL if they need it, but it's always trick to change existing behavior.

I think the best would be to add a `suffix` parameter, that would be ".git" by default, and allow `nil` to disable the feature ?

PR #403, opened 2024-03-08
add option to disable the .git suffix

annaborn:
Hello Sylvain @doudou !

Here is the commit to the issue https://github.com/rock-core/autoproj/issues/400.
I thought, there is only one possible suffix for git repositories, .git. So I have modified your advice from the issue, and added the option to disable the suffix without the possibility to modify the suffix itself.
So at the end, the user can leave everything how it is, and per default .git will be added to the repos. Or use add_suffix: false to get the repository urls without .git ending. 

I tested it on my local setup. Where I had two git server configurations: one without add_suffix option and one with add_suffix: false.


annaborn:
@doudou Sylvain, could you please check the Pull Request?

PR #406, opened 2024-07-04
Fix yaml safe load

g-arjones:
Passing whitelisted classes to `YAML$safe_load` as a positional argument has been deprecated in Psych v3.1.0 (Ruby v2.7.8) and completely removed in v4.0.3 (Ruby v3.1.0)


g-arjones:
@2maz Could you have a look at this, please?

2maz (approved):
LGTM. 

I see, however, that checks fail due to other reasons.

g-arjones:
It looks like there's actually something wrong with the bundle versioning code. Unfortunately, I can't have a deeper look right now.

Thanks for the review.

PR #413, opened 2025-07-15
allow installing ruby shims in a separate prefix

g-arjones:
CMake automatically adds the parent folder of each path in `PATH` to its search paths, (partially) defeating the purpose of the setting.

### DIFF
--- a/lib/autoproj/configuration.rb
+++ b/lib/autoproj/configuration.rb
@@ -436,6 +436,22 @@ module Autoproj
             set("separate_prefixes", flag, true)
         end
 
+        # Returns true if Ruby shims should be installed in a separate prefix
+        #
+        # The default is false (disabled)
+        #
+        # @return [Boolean]
+        def isolate_ruby_shims?
+            get("isolate_ruby_shims", false)
+        end
+
+        # Controls whether Ruby shims should be in a separate prefix
+        #
+        # @see isolate_shims?
+        def isolate_ruby_shims=(flag)
+            set("isolate_ruby_shims", flag, true)
+        end
+
         # Returns true if packages and prefixes should be auto-generated, based
         # on the SHA of the package names. This is meant to be used for build
         # services that want to check that dependencies are properly set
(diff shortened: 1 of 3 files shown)


g-arjones:
To me, it would also be fine to _always_ install to `prefix/autoproj/bin` regardless of `config.separate_prefixes?`

g-arjones:
> Mixing different versions of python or ruby in a single workspace would be very fragile.

That's not the goal. The issue I'm trying to fix is that cmake will add the parent folder of each folder in `PATH` to its internal search paths. This means that `#{Autoproj.prefix}/install` always ends up in cmake's search paths and it will sometimes use or find stuff that it should not when using `separate_prefix = true` because cmake does a lot of heuristics when it is looking for things.  If the shims are in `#{Autoproj.prefix}/install/autoproj/bin` instead then ``#{Autoproj.prefix}/install/autoproj` will be added (which is fine, because that folder will only have the shims anyway).

> in any case, this patch would break a lot of stuff, no ? 

I don't see why, and it didn't in my tests but I can put it behind a configuration flag if you prefer.

> You would at least need to add the new path to Autoproj's internal path, and make sure it ends up where it should in env.sh.

[It's there already (the second line following the last line of the patch)](https://github.com/rock-core/autoproj/blob/8f351e6b69e04ca9385fad85e17dae2137d5e461/lib/autoproj/workspace.rb#L417)

g-arjones:
> But why not move it in both cases (separate_prefixes and normal) ? 

I'm fine with that.

> Wouldn't the same issue apply with separate_prefixes = false ?

It would, but that's kind of expected in that case because all packages share the same prefix anyway.

I will change the patch to have the shims always in a separate prefix, regardless of `config.separate_prefixes`, and have it behind a flag.

@doudou Could you please have a look at https://github.com/rock-core/autoproj/pull/395 ? That one has been open since 2023.

g-arjones:
@doudou Done!

doudou (approved):

PR #414, opened 2025-08-06
Bundler 2.7 compatibility

g-arjones:
On top of #395 

See https://bundler.io/changelog.html#270-2025-07-16
and https://github.com/rubygems/rubygems/pull/8345

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        ruby-version: ["3.0", "2.7"]
+        ruby-version: ["3.2", "3.1", "2.7"]
     continue-on-error: true
     steps:
     - uses: actions/checkout@v2
--- a/bin/autoproj_bootstrap
+++ b/bin/autoproj_bootstrap
@@ -85,7 +85,9 @@ module Autoproj
                 env["RUBYLIB"] = []
                 env["PATH"] = self.class.sanitize_env(ENV["PATH"] || "")
                 env["BUNDLE_GEMFILE"] = []
-
+                env["BUNDLER_VERSION"] = []
+                env["BUNDLER_SETUP"] = []
+                env["BUNDLE_BIN_PATH"] = []
                 load_config
 
                 if config["ruby_executable"] != Gem.ruby
@@ -264,7 +266,7 @@ module Autoproj
             # @param [String] autoproj_version a constraint on the autoproj version
             #   that should be used
             # @return [String]
-            def default_gemfile_contents(autoproj_version = ">= 2.17.0")
+            def default_gemfile_contents(autoproj_version = ">= 2.18.0")
                 ["source \"#{gem_source}\"",
                  "ruby \"#{RUBY_VERSION}\" if respond_to?(:ruby)",
                  "gem \"autoproj\", \"#{autoproj_version}\""].join("\n")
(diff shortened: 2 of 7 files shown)


g-arjones:
Alternatively, we could also have something like:

```ruby
            def write_bundle_shim(shims_path, bundler_version: nil)
                script = <<~INSTALLER
                    require 'rubygems/installer'
                    spec = Gem::Specification.find_by_name("bundler", '#{bundler_version || '>= 0'}')
                    installer = Gem::Installer.for_spec(spec)
                    installer.generate_bin_script("bundle", '#{shims_path}')
                INSTALLER

                pid = fork { exec(env_for_child, Gem.ruby, "-e", script) }
                Process.wait(pid)

                raise "Could not generate 'bundle' binstub" unless $CHILD_STATUS.success?
            end
```

Instead of hardcoding the contents of the shim... @doudou thoughts?

g-arjones:
Tagging @dbcesar since you mentioned you would like to have a look

doudou:
> Alternatively, we could also have something like:
> 
> ```ruby
> def write_bundle_shim(shims_path, bundler_version: nil)
>     script = <<~INSTALLER
>         require 'rubygems/installer'
>         spec = Gem::Specification.find_by_name("bundler", '#{bundler_version || '>= 0'}')
>         installer = Gem::Installer.for_spec(spec)
>         installer.generate_bin_script("bundle", '#{shims_path}')
>     INSTALLER
> 
>     pid = fork { exec(env_for_child, Gem.ruby, "-e", script) }
>     Process.wait(pid)
> 
>     raise "Could not generate 'bundle' binstub" unless $CHILD_STATUS.success?
> end
> ```
> 
> Instead of hardcoding the contents of the shim... @doudou thoughts?

It feels better.

The fork/exec/wait is done by `system`, which returns false if the command failed.

doudou:
Hey ... I might be a tiny bit lost here.

`bundle` is not generating a binstub for `bundle`, but `gem` is, and we install bundler with `gem`. I don't think this is the right fix, actually.

g-arjones:
> I guess we should be picking bundle from $GEM_HOME/bin instead of from our binstub path.

Yes, my initial plan was to store the path to the gem-generated stub in a `bundle_executable` config setting and then use it instead of the autoproj-generated one in `BundlerManager`. However, in the end, I figured the current patch was better with regard to backward compatibility.

In hindsight, maybe we could have a "minimal" shim that just exec's the stub generated by gem. Similar to what we do with `.autoproj/bin/ruby`. This would be backward-compatible and avoid having to write the stub ourselves. What do you think?

doudou:
> > I guess we should be picking bundle from $GEM_HOME/bin instead of from our binstub path.
> 
> Yes, my initial plan was to store the path to the gem-generated stub in a `bundle_executable` config setting and then use it instead of the autoproj-generated one in `BundlerManager`. However, in the end, I figured the current patch was better with regard to backward compatibility.
> 
> In hindsight, maybe we could have a "minimal" shim that just exec's the stub generated by gem. Similar to what we do with `.autoproj/bin/ruby`. This would be backward-compatible and avoid having to write the stub ourselves. What do you think?

Sounds good. The bundler/rubygem's people plan is definitely to have people use the gem-generated stub, better follow their plan ;-)

g-arjones:
@doudou Done!

PR #416, opened 2025-08-13
Improve vscode integration

g-arjones:
The rebornix.ruby extension has been deprecated for a while. This replaces it with `shopify.ruby-lsp`.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,20 @@
+{
+    "name": "autoproj",
+    "image": "ubuntu:24.04",
+    "features": {
+        "ghcr.io/devcontainers/features/common-utils:2": {},
+        "ghcr.io/devcontainers-extra/features/apt-get-packages:1": {
+            "packages": "bash-completion,sudo,libyaml-dev,ruby-dev,build-essential,git,python3-pip"
+        }
+    },
+    "remoteUser": "ubuntu",
+    "waitFor": "postCreateCommand",
+    "postCreateCommand": "./.devcontainer/post-create.sh",
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "Shopify.ruby-extensions-pack"
+            ]
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


planthaber:
> > Unfortunately I'm still stuck on ubuntu 20.04 ... lsp does not work there.
> 
> Ever tried devcontainers ? Pretty handy.

They are, in dfki we use this script set that handles the the docker parameters using a settings file: https://github.com/dfki-ric/docker_image_development

We even use it to run on the robots, for keeping our old stacks runnable after upgrading the os on the robot.
It also makes sure the robot has exact the same ubuntu version as the dev PCs, as they use the same docker image.
And we can bake a dev state into a docker image to keep it runnable while going forward in development.

Features are accessible through the settings.bash file of the scrips:

* XServer - Graphical programs
  * 3d acceleration (only for nvidia cards)
  * GUIs on remote mashines
    *  Ssh -X tunnels to containers on remote machines (use ssh -X to connect a machine whre a GUI runs in a docker)
    *  Or automated launch of an xpra server in the container
* Use the hosts ssh-agent in the container (no need to copy, mount or create an key in the container for e.g. ssh checkouts)
* Creating images from workspaces (e.g. to save a runnable version of a  successful demo)
* Shares the settings/image names through a common git repo, 
* ccache and icecc support (compiler cache and distributed compilation)
* docker in docker
* Mouting the complete container filesystem on the host (needs sudo on startup)
  * To be used for profiling with perf/hotspot on the host

g-arjones:
@planthaber  Sounds cool, but I think we are talking about different things. 

I'm talking about devcontainers(TM): See [containers.dev](https://containers.dev/). The configuration file provided in this PR would allow building a devcontainer specifically for developing autoproj.

planthaber:
> @planthaber Sounds cool, but I think we are talking about different things.
> 
> I'm talking about devcontainers(TM): See [containers.dev](https://containers.dev/). The configuration file provided in this PR would allow building a devcontainer specifically for developing autoproj.

Got the devcontainers(TM) part, but didn't realize that it is part of this PR. Thought it was a suggestion to test with another OS. (which can be doen in both ways). Anyways, having the devcontainers config in here seems beneficial.

PR #417, opened 2025-08-15
Fix and extend dependency condition handling in PackageManifest

gabrielfpacheco:
This PR addresses two potential issues in `PackageManifest` dependency evaluation:

1. Handling of missing configuration variables.
   - Previously, if a configuration variable did not exist, evaluating a dependency condition could raise an error. The change ensures that missing variables are treated as empty strings to avoid breaking the evaluation.

2. Boolean condition evaluation.
   - Conditions using boolean values (e.g., `$FOO==true` or `$BAR==false`) were not interpreted correctly and would fail. The update normalizes these evaluations so they behave as expected.


g-arjones (approved):

gabrielfpacheco:
Hey @pierrewillenbrockdfki,

I noticed that the following test (introduced in https://github.com/rock-core/autoproj/pull/382) is failing in CI. It seems the test expects an exception when a configuration variable does not exist:

```ruby
            it "raises an Autoproj::ConfigError if an unknown configuration is found" do
                assert_raises(Autoproj::ConfigError) do
                    PackageManifest::Loader.expand_configuration_variable(
                        "$BAR", @config
                    )
                end
            end
```

Could you confirm if this is the intended behavior? If it is, I can adjust this patch to focus solely on fixing the boolean-to-string conversion issue, without changing how missing configurations are handled.

g-arjones:
I'm really not sure... Doesn't this defeat the purpose of configuration settings having default values in the first place?

PR #427, opened 2025-12-30
feat: update the key management scheme to be compatible with recent Debian key management

doudou:
apt-key is deprecated. One is now supposed to save the keys separately in
/etc/apt/keyrings and refer to them with the signed-by attribute.

This commit updates the existing scheme to allow for named keys, that is
the 'keys' entries in osrepos can have a 'name' entry, and the 'repo' entries
should have a 'key' entry with the name of that key

If a key has a name, it is saved in /etc/apt/keyrings instead of being passed
to apt-key. If a repo has a 'key' entry, it gets the corresponding signed-by
attribute.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,18 @@ gem "rubygems-server" unless RUBY_VERSION < "3"
 group :dev do
     gem "rubocop", "~> 1.28.0"
     gem "rubocop-rock"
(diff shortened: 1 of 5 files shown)


g-arjones:
Cool! Tests are failing though...

PR #431, opened 2026-02-18
Set 'BUNDLE_LOCKFILE' to user's lock file

g-arjones:
This is also required on Bundler 4.0.4+

### DIFF
--- a/bin/autoproj_bootstrap
+++ b/bin/autoproj_bootstrap
@@ -450,6 +450,7 @@ module Autoproj
                 contents = <<~BUNDLERSHIM
                     #!#{Gem.ruby}
                     ENV['BUNDLE_GEMFILE'] ||= '#{autoproj_gemfile_path}'
+                    ENV['BUNDLE_LOCKFILE'] ||= '#{autoproj_gemfile_path}.lock'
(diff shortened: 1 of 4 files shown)


g-arjones:
@doudou Implemented this on top of your `bundler-4` branch, but feel free to change the base if you want.

doudou:
The default lockfile is not GEMFILE.lock anymore ?

g-arjones:
It is, but `require 'bundler/setup'` now sets `BUNDLE_LOCKFILE` explicitly in the shims. Then, when `BundlerManager` initializes the environment and sets `BUNDLE_GEMFILE` to `prefix/gems/Gemfile`, we end up with one pointing to the user Gemfile and the other to the internal lock file.

To make matters even worse, if we only set `BUNDLE_LOCKFILE` in `BundlerManager#initialize_environment` and not in the shims, when you source `env.sh`  you will have `BUNDLE_LOCKFILE` pointing to the user's lock file and `BUNDLE_GEMFILE`  pointing to the internal gem file (set in the shim) which will cause bundler to refuse to load completely due to missing gems in the lock file.

doudou (approved):

PR #433, opened 2026-04-07
Allow newer utilrb versions

chhtz:
The current dependency

    "utilrb", "~> 3.0.0", ">= 3.0.0"

only allows utilrb versions 3.0.x (which is why autoproj still installs version 3.0.1, even though version 3.1.0 exists)

The `">= 3.0"` is actually implied by `"~> 3.0"` and could be removed, if I understand this correctly (similar for other dependencies).


chhtz:
Unit tests currently fail, since utilrb 3.1.0 is also incompatible with newest facets version.

cf: https://github.com/orocos-toolchain/utilrb/pull/49

doudou (approved):

g-arjones:
@doudou Can we merge this and release utilrb with https://github.com/rock-core/tools-utilrb/pull/5, please?

PR #434, opened 2026-05-30
fix: make sure BUNDLE_GEMFILE and BUNDLE_LOCKFILE are consistent when invoking bundler

g-arjones:
This fixes a bug introduced by https://github.com/rock-core/autoproj/pull/431

In some situations (e.g, when calling autoproj using the shim and if `install/gems/Gemfile.lock` was missing), there would be a mismatch between `BUNDLE_GEMFILE` and `BUNDLE_LOCKFILE` in `BundlerManager`.


g-arjones:
The test suite is broken until https://github.com/rock-core/autoproj/pull/433 is resolved

doudou (approved):

